PR #65, opened 2017-09-28
Rename GameObjects and allow rectangular sprites.

pathunstrom:


ppb/abc.py
```diff
@@ -39,26 +40,28 @@ def run(self):
...
-    def image(self) -> 'Canvas':
+    @position.setter
+    @abc.abstractmethod
+    def position(self, value):
         raise NotImplementedError
 
-    @property
-    def scene(self) -> 'Scene':
+    @abc.abstractmethod
+    def __contains__(self, item):
```
astronouth7303: All GameObjects are containers?

pathunstrom: Not in the traditional sense. This one probably counts as a hack. Also probably shouldn't be in this PR.

Issue #64, opened 2017-09-27
Add docstrings to engine.GameEngine


pathunstrom:

ali-alsabbah:
I'm interested in doing this if it's still open, but I might need a little guidance

pathunstrom:
Basics of docstrings is in [pep 257](https://www.python.org/dev/peps/pep-0257/).

So, for an example of what I'm looking for as "best": Check out [the scenes module](https://github.com/ppb/pursuedpybear/blob/master/ppb/scenes.py) for a sample.

If you just want the commit, I'll take even basic docstrings.

pathunstrom:
Resolved with #423

PR #69, opened 2017-09-28
Created docstrings for engine.GameEngine

Fedorsturov:
Resolved issue #64

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -8,6 +8,19 @@ from ppb.abc import Engine, Scene
 class GameEngine(Engine):
 
     def __init__(self, first_scene: Type, **kwargs):
+        """
+        Game Engine
+
+        first_scene -- the scene to be activated first
+
+        delta_time -- time between frames
+
+        resolution -- window (w, h) tuple
+
+        log_level -- log level for all messages
+
+        flags and depth are unused
+        """
         super(GameEngine, self).__init__()
 
         # Engine Configuration
@@ -51,6 +64,9 @@ class GameEngine(Engine):
             self.activate(next_scene)
 
     def run(self):
+        """
+        Starts main loop
+        """
         self.start()
         while self.running:
             time.sleep(.0000000001)
@@ -73,12 +89,20 @@ class GameEngine(Engine):
 
     @property
     def current_scene(self):
+        """
+        Get current scene
+        """
         try:
             return self.scenes[-1]
         except IndexError:
             return None
 
     def activate(self, next_scene: dict):
+        """
+        Activates a specific scene
+
+        next_scene -- scene to be activated
+        """
         scene = next_scene["scene_class"]
         if scene is None:
             return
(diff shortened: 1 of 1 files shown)


ppb/engine.py
```diff
@@ -8,6 +8,19 @@
 class GameEngine(Engine):
```
pathunstrom: Missing class docstring.

ppb/engine.py
```diff
@@ -8,6 +8,19 @@
...
 
     def __init__(self, first_scene: Type, **kwargs):
+        """
+        Game Engine
```
pathunstrom: This isn't documentation, and echoes the name of the class.

We can change this to "Create GameEngine" or ignore it since __init__'s usage is well known.

ppb/engine.py
```diff
@@ -8,6 +8,19 @@
...
+
+        first_scene -- the scene to be activated first
+
+        delta_time -- time between frames
+
+        resolution -- window (w, h) tuple
+
+        log_level -- log level for all messages
+
+        flags and depth are unused
```
pathunstrom: This is false, both flags and depth are passed to `pygame.display.set_mode()`.

ppb/engine.py
```diff
@@ -8,6 +8,19 @@
...
+        """
+        Game Engine
+
+        first_scene -- the scene to be activated first
+
+        delta_time -- time between frames
+
+        resolution -- window (w, h) tuple
+
+        log_level -- log level for all messages
```
pathunstrom: log level is the minimum log level for printing to the console. It defaults to logging.WARNING to minimize end user noise, but the keyword exists specifically for developers to switch down to INFO or DEBUG for debugging.

ppb/engine.py
```diff
@@ -73,12 +89,20 @@ def run(self):
...
+        Get current scene
+        """
         try:
             return self.scenes[-1]
         except IndexError:
             return None
 
     def activate(self, next_scene: dict):
+        """
+        Activates a specific scene
```
pathunstrom: Should be written as a command. (Drop the "s" on "activates"). Specific is probably unneeded.

ppb/engine.py
```diff
@@ -73,12 +89,20 @@ def run(self):
...
+
+        next_scene -- scene to be activated
```
pathunstrom: I'd like the expected structure of next_scene to be documented here:

* scene: a Type
* args: a list or tuple
* kwargs: a dictionary

This is important for end users, because this is one of the primary interfaces for scenes to replace themselves.

ppb/engine.py
```diff
@@ -87,5 +111,8 @@ def activate(self, next_scene: dict):
         self.scenes.append(scene(self, *args, **kwargs))
 
     def update_input(self):
+        """
+        Updates mouse state
```
pathunstrom: Phrase as a command, also, upate will also update the keyboard object in the future, so "update input devices" is a more generic way to phrase this.

ppb/engine.py
```diff
@@ -8,6 +8,19 @@
 class GameEngine(Engine):
 
     def __init__(self, first_scene: Type, **kwargs):
+        """
+        Game Engine
+
+        first_scene -- the scene to be activated first
+
+        delta_time -- time between frames
```
pathunstrom: "time between simulation frames" would be more clear.

ppb/engine.py
pathunstrom: I've changed my mind on this. I won't block this PR based on a docstring missing.

pathunstrom:
Closed due to obsolete.

Issue #53, opened 2017-08-16
Use new Keyword only parameters in GameEngine


pathunstrom:
Replace the **kwargs with [keyword only arguments](https://github.com/pathunstrom/pursuedpybear/commit/8fa08556daa61af7932fb5e90cb3ba4f13e0bd86##commitcomment-23681597).

imjennyli:
@pathunstrom I'd like to take this on, would that be alright?

Issue #54, opened 2017-08-16
Keyword only parameters in BaseScene


pathunstrom:

schamp:
I can take a stab at this.  It looks like the idea is to replace kwargs with 'background_color' as a keyword argument with a default value?

pathunstrom:
Not necessarily replace kwargs, but to take the get() and turn it into a [keyword only argument](https://www.python.org/dev/peps/pep-3102/). Because of the nature of the API and the tendency to subclass APIs, I'd like to allow keywords to be passed arbitrarily to prevent breaking changes in the future.

schamp:
That makes sense, I think I got it.

PR #70, opened 2017-09-29
Replace **kwargs with keyword only arguments (#53)

droudy:


pathunstrom:
This PR does not actually do the work required and mixes work:

1. The issue asked for keyword only arguments. This has just added keyword arguments with defaults, both issues want [keyword only parameters](https://www.python.org/dev/peps/pep-3102/).
2. This attempts to resolve work for more than one issue ( #53 , #54 ), please limit PRs to one issue each.
3. This breaks an interface in Canvas, which is intended to be usable as a mix-in, and thus requires **kwargs.

For these reasons, I'm closing this PR. Please feel free to work on these issues in separate PRs, one for issue #53 and one for issue #54.

PR #73, opened 2017-10-04
change GameEngine to be keyword-only

imjennyli:
Fixes https://github.com/ppb/pursuedpybear/issues/53

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -7,15 +7,15 @@ from ppb.abc import Engine, Scene
 
 class GameEngine(Engine):
 
-    def __init__(self, first_scene: Type, *, delta_time=None, resolution=None, flags=None, depth=None, log_level=None):
+    def __init__(self, first_scene: Type, *, delta_time=0.016, resolution=(600, 400), flags=0, depth=0, log_level=logging.WARNING, **kwargs):
         super(GameEngine, self).__init__()
 
         # Engine Configuration
-        self.delta_time = delta_time or 0.016
-        self.resolution = resolution or (600, 400)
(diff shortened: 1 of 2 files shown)


ppb/engine.py
```diff
@@ -7,15 +7,15 @@
 
 class GameEngine(Engine):
 
-    def __init__(self, first_scene: Type, **kwargs):
+    def __init__(self, first_scene: Type, *, delta_time=None, resolution=None, flags=None, depth=None, log_level=None):
```
pathunstrom: Please add `**kwargs` as well. If we remove one of these keywords in the future, I'd like to prevent breaking existing code, so going to operate that `**kwargs` is always there.

ppb/engine.py
```diff
@@ -7,15 +7,15 @@
...
         super(GameEngine, self).__init__()
 
         # Engine Configuration
-        self.delta_time = kwargs.get("delta_time", 0.016)
-        self.resolution = kwargs.get("resolution", (600, 400))
-        self.flags = kwargs.get("flags", 0)
-        self.depth = kwargs.get("depth", 0)
-        self.log_level = kwargs.get("log_level", logging.WARNING)
+        self.delta_time = delta_time or 0.016
```
pathunstrom: There's no reason to use the None sentinel in any of these cases: None of these defaults are mutable and it'll read more cleanly if they're in the signature.

Please switch the defaults from `None` to the respective values.

ppb/engine.py
imjennyli: Ahh good point

PR #74, opened 2017-10-06
Change BaseScene background_color arg to be keyword-only (Fixes #54)

### LINKED ISSUE #54: Keyword only parameters in BaseScene

schamp:
I struggled to run the tests, are there instructions somewhere for setting up the environment, running the tests, etc.?


pathunstrom:
There aren't currently instructions, but `pip install -e .` from the top directory and then `python -m unittest` in the test directory should work. The coverage is less than stellar.

schamp:
Got it, thanks, that worked for me.  I'll write a separate issue and PR for putting those unit test instructions in the CONTRIBUTING.md

PR #93, opened 2018-05-14
Fix base scene rendering

pathunstrom:


ppb/scenes.py
```diff
@@ -17,20 +20,21 @@ def __init__(self, engine, *, background_color=(0, 0, 55),
         self.background = engine.display.copy()
         self.background.fill(self.background_color)
         if container_class is None:
-            container_class = GameObjectContainer
+            container_class = GameObjectCollection
```
astronouth7303: Maybe make this signature default?

Issue #99, opened 2018-09-02
Camera Sub System


pathunstrom:
Adds a camera object for other sub systems to rely on.

PR #102, opened 2018-09-02
New Camera Class

pathunstrom:
Provides three basic concepts: The Camera object itself, the viewport which is the space in the rendered window it applies to, and the frame, which is the area it sees in the game world.

Concerns #99

### DIFF
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -1,10 +1,8 @@
 from ppb.vector import Vector
 from ppb.engine import GameEngine
 
-from ppb.sprites import BaseSprite
-# Needs to come after BaseSprite because it technically depends on it.
 from ppb.scenes import BaseScene
-
+from ppb.sprites import BaseSprite
 
 
 def run(starting_scene=BaseScene, **kwargs):
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -1,8 +1,10 @@
 from ppb.vector import Vector
 from ppb.engine import GameEngine
 
-from ppb.scenes import BaseScene
 from ppb.sprites import BaseSprite
+# Needs to come after BaseSprite because it technically depends on it.
+from ppb.scenes import BaseScene
+
 
 
 def run(starting_scene=BaseScene, **kwargs):
--- a/ppb/camera.py
+++ b/ppb/camera.py
@@ -56,6 +56,10 @@ class Camera(BaseSprite):
                 self.frame_bottom >= sprite.top
                 )
 
-    def translate_to_game_space(self, point:Vector) -> Vector:
+    def translate_to_frame(self, point:Vector) -> Vector:
         offset = (point - self.viewport_offset) * (1/self.pixel_ratio)
         return self.position + offset
+
+    def translate_to_viewport(self, point:Vector) -> Vector:
+        offset = (point - self.position) * self.pixel_ratio
+        return self.viewport_offset + offset
--- a/ppb/camera.py
+++ b/ppb/camera.py
@@ -10,9 +10,13 @@ class Camera(BaseSprite):
         """
 
         viewport: A container of origin x, origin y, width, and
-                  height in pixel space.
-        pixel_ratio: An number defining the pixel to game space ratio.
-                     Effectively a "zoom".
+                  height. The origin is the top left point of the viewport
+                  measured from the top left point of the window or screen.
+                  The width and height are the raw pixel measurements of the
+                  viewport.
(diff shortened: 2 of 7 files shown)


ppb/__init__.py
```diff
@@ -1,8 +1,10 @@
 from ppb.vector import Vector
 from ppb.engine import GameEngine
 
-from ppb.scenes import BaseScene
 from ppb.sprites import BaseSprite
+# Needs to come after BaseSprite because it technically depends on it.
+from ppb.scenes import BaseScene
```
astronouth7303: I would have thought that module resolution would have handled this? But I'm guessing this is here because you had a bug.

ppb/camera.py
```diff
@@ -0,0 +1,67 @@
...
+    image = DoNotRender
+
+    def __init__(self, viewport=(0, 0, 800, 600), pixel_ratio=80):
+        """
+
+        viewport: A container of origin x, origin y, width, and
+                  height in pixel space.
+        pixel_ratio: An number defining the pixel to game space ratio.
+                     Effectively a "zoom".
+        """
```
astronouth7303: Soooo.... viewport width/height divided by the pixel ratio is the actual window width/height in screen pixels? It should probably not be labeled "pixel space".

Also, which corner is the origin again?

ppb/camera.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def in_frame(self, sprite: BaseSprite):
+        return (self.frame_left <= sprite.right and
+                self.frame_right >= sprite.left and
+                self.frame_top <= sprite.bottom and
+                self.frame_bottom >= sprite.top
+                )
+
+    def translate_to_frame(self, point:Vector) -> Vector:
+        offset = (point - self.viewport_offset) * (1/self.pixel_ratio)
```
astronouth7303: That's... an odd way to express that equation.

ppb/flags.py
```diff
@@ -0,0 +1,19 @@
...
+Use like None:
+    * compare against using `is`
+    * Don't instantiate a new instance.
+"""
+
+
+class Flag:
+    """Inherit from Flag to make a simple class flag."""
+    def __new__(cls, *args, **kwargs):
+        return cls
```
astronouth7303: What happens if you actually instantiate one? (I keep thinking something odd happens with `__init__()` if you return an existing object from `__new__()`.)

Would it be simpler/easier to just raise a `TypeError`?

ppb/__init__.py
pathunstrom: Short version: Circular import problems. Should maybe set a policy to not use the `__init__.py` exports in internal modules and to use the modules directly.

ppb/camera.py
pathunstrom: Opposite: Viewport width x height == actual pixel width and height of the viewport. Divide by pixel ratio to get to game space.

Top left is the origin. Definitely should not that.

ppb/camera.py
pathunstrom: What do you think is a better way to write it?

ppb/flags.py
pathunstrom:
```
>>> flag = Flag(1, 2, 3)
>>> flag is Flag
True
```

We could also raise a TypeError. This is definitely not an API I'm set on quite yet.

astronouth7303:
Hm. Should scenes have a helper method to get the camera?

pathunstrom:
Regarding multiple camera: Currently, no, it's not meaningful. When it's decided to do that work (very firmly _not_ before 1.0) a Camera is a translation between its viewport and it's frame. So rendering a camera would actually go through and render a frame that would then added to the window where the viewport is.

ppb/__init__.py
astronouth7303: That's a good idea.

ppb/camera.py
astronouth7303: `(point - self.viewport_offset) / self.pixel_ratio`

pathunstrom: Ah, yeah, that's because `Vector` doesn't support division.

Not sure how meaningful it would be to include that in its API.

pathunstrom:
There's a bug regarding the default resolution. Probably need to find a way to pass a reference to the default Camera system.

ppb/camera.py
astronouth7303: Oh! 

Scalar division makes sense? As the shorter equivalent to `v * (1 / x)`.

PR #107, opened 2018-09-03
Mouse system

pathunstrom:


ppb/mouse.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    def __init__(self, press_function: Callable=None,
+                 move_function: Callable=None, move_to_function: Callable=None):
+        self.position = Vector(0, 0)
+        self.screen_position = Vector(0, 0)
+        self.buttons = [False, False, False]
+
+        self.press_function = press_function or _not_implemented
+        self.move_function = move_function or _not_implemented
+        self.move_to_function = move_to_function or _not_implemented
```
ironfroggy: if you did `partial(_not_implemented, "...")` with the error messages you use below you could avoid the wrapper methods and just assign to `press()` directly here

ppb/mouse.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @property
+    def left_button(self) -> bool:
+        return self.buttons[0]
+
+    @property
+    def right_button(self) -> bool:
+        return self.buttons[2]
```
ironfroggy: If i was using this I'd probably start to wish these were shorter, just `left` and `right`

ppb/mouse.py
```diff
@@ -0,0 +1,53 @@
...
+        except NotImplementedError:
+            raise FeatureNotProvided("Your hardware library did not include a Mouse.move_to function.")
+
+
+class FeatureNotProvided(Exception):
+    pass
+
+
+def _not_implemented(*_):
+    raise NotImplementedError
```
ironfroggy: if this is _only_ used to catch and reraise something else, why not just raise `FeatureNotProvided` directly?

On a related note, should that exception subclass from `NotImplementedError`?

ppb/systems.py
```diff
@@ -26,9 +33,19 @@ class PygameEventPoller(System):
...
-    event_map = {
-        pygame.QUIT: events.Quit,
-    }
+    event_map = None
+
+    def __new__(cls, *args, **kwargs):
+        if cls.event_map is None:
+            cls.event_map = {
+                pygame.QUIT: "quit",
+                pygame.MOUSEMOTION: "mouse_motion"
```
ironfroggy: nitpick! ending every collection entry with a comma, including the last one, makes for cleaner diffs later (so you don't need to edit a line just to add a comma)

ppb/systems.py
```diff
@@ -40,7 +57,61 @@ def on_update(self, update, signal):
         for pygame_event in pygame.event.get():
             ppb_event = self.event_map.get(pygame_event.type)
             if ppb_event is not None:
-                signal(ppb_event())
+                signal(getattr(self, ppb_event)(pygame_event, update.scene))
```
ironfroggy: If you extract the `getattr()` call you can catch any `AttributeError` and give a more concise error if an event doesn't have a matching method.

ppb/systems.py
```diff
@@ -40,7 +57,61 @@ def on_update(self, update, signal):
...
+
+    def quit(self, event, scene):
+        return events.Quit()
+
+    def mouse_motion(self, event, scene):
+        screen_position = Vector(*event.pos)
+        camera = list(scene.get(tag="main_camera"))[0]
+        game_position = camera.translate_to_frame(screen_position)
+        delta = Vector(*event.rel) * (1/camera.pixel_ratio)
```
ironfroggy: Would be a good place for a comment about what the `delta` is

ppb/systems.py
```diff
@@ -40,7 +57,61 @@ def on_update(self, update, signal):
...
+
+    def __init__(self, *, engine, resolution=default_resolution, **kwargs):
+        self.mouse = Mouse()
+        self.offset = Vector(-0.5 * resolution[0],
+                             -0.5 * resolution[1])
+        engine.register(events.Update, "mouse", self.mouse)
+
+    def activate(self, engine):
+        """Sync mouse with hardware."""
+        buttons = self.get_hardware_buttons()
```
ironfroggy: Does this even return a different length for `buttons` and, if so, what would happen?

ppb/mouse.py
pathunstrom: Love this. Definitely making both of these changes.

ppb/mouse.py
pathunstrom: By the same extension, perhaps these should be properties.

ppb/mouse.py
pathunstrom: You mean as the second half of the assignment here?

If so, I can dig it, I'll throw that together and see how it works.

ppb/systems.py
pathunstrom: Big catch is that if the event exists in the lookup dictionary, it should exist on the system and failure here is definitely a thing to catch before it's in a user's hands.

ppb/systems.py
pathunstrom: The big thing I need to figure out is if delta or "rel" is more understandable.

ppb/systems.py
pathunstrom: This is a fun one: It's going to be hardware lib (and probably actual hardware) dependent.

In a given run time? No, it shouldn't.

Between runtimes: Possibly.

That said, it should just work since it will be an array like object and the mouse can handle that.

PR #126, opened 2018-09-25
Refactor Logging

astronouth7303:
Handle logging better.

This also includes a `LoggingMixin` that handles the (admittedly small) boiler plate of getting & caching a logger.

### DIFF
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -4,7 +4,10 @@ from ppb.engine import GameEngine
 from ppb.scenes import BaseScene
 from ppb.sprites import BaseSprite
 
+import logging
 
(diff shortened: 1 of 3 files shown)


astronouth7303:
I'm also considering Fixing `events.py` to use this? That will move the data currently logged at `game.events`.

pathunstrom:
I like that idea, but do we want EventMixIn to inherit from LoggerMixIn?

astronouth7303:
ummm.... not really. That sounds like "yeah, this is implicitly on all your objects forever", which isn't super great.

astronouth7303:
Oh, and it wouldn't do the right thing on a mixin anyway, so nevermind!

PR #133, opened 2018-09-29
Move scene defaults to class level

astronouth7303:
Keep scene defaults in the class definition, so they can be simply overriden:

```python
class MainScene(ppb.BaseScene):
    background_color = (0, 0, 0)
```


astronouth7303:
I would like to move `pixel_ratio` to the class, too, but also don't set up attribute synchronization.

pathunstrom:
So, I actually worry about this: By moving these to class attributess, we lose the type hinting in the signature, which makes discoverability for a new user in an IDE much harder.

astronouth7303:
You can annotate class attributes. The problem is that we loose annotation on the signature itself, which is... :sigh:

Without additional tooling, I think we have to choose between convenient code (including convenient extensibility) or 100% correctly annotated, no cases left out.

Maybe we can do some metaprogramming that'll generate the right signature for subclasses?

How does dataclasses and attrs handle this? Do they?

ppb/scenes.py
```diff
@@ -96,16 +96,20 @@ def remove(self, game_object: Hashable) -> None:
 
 
 class BaseScene(Scene, EventMixin):
+    # Background color, in RGB, each channel is 0-255
+    background_color: Sequence[int] = (0, 0, 100)
 
-    def __init__(self, engine, *, background_color: Sequence[int]=(0, 0, 100),
+    def __init__(self, engine, *,
                  container_class: Type=GameObjectCollection,
```
pathunstrom: This can probably be included.

ppb/scenes.py
```diff
@@ -96,16 +96,20 @@ def remove(self, game_object: Hashable) -> None:
...
                  set_up: Callable=None, pixel_ratio: Union[int, float]=80,
                  **kwargs):
         super().__init__(engine)
-        self.background_color = background_color
-        self.background = None
+        for k, v in kwargs.items():
```
pathunstrom: This has taken some time, but I'm super for this now that I think about it. It means you don't even need to subclass if you don't want to.

astronouth7303:
Should I do the same for sprites?

PR #137, opened 2018-09-30
Add support for setting the window title

astronouth7303:
Just a little thing that helps.

### DIFF
--- a/ppb/systems.py
+++ b/ppb/systems.py
@@ -125,14 +125,17 @@ class Quitter(System):
 
 class Renderer(System):
 
-    def __init__(self, resolution=default_resolution, **kwargs):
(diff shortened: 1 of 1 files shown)


astronouth7303:
(Sorry, I'm not sure how to test this.)

ppb/systems.py
```diff
@@ -125,14 +125,17 @@ def activate(self, engine):
 
 class Renderer(System):
 
-    def __init__(self, resolution=default_resolution, **kwargs):
+    def __init__(self, resolution=default_resolution, window_title=None, **kwargs):
```
pathunstrom: Instead of `None`, let's default to `"PursuedPyBear"`. Bonus, we don't need the None check at L137.

Issue #156, opened 2018-10-04
Keyboard Button Down


pathunstrom:
We need a Keyboard button down event.

This card is complete when:

- [ ] A new event type is defined
- [ ] The event poller raises one as expected.

PR #160, opened 2018-10-05
[WIP] Update examples

astronouth7303:
Update example games

### DIFF
--- a/examples/framecount.py
+++ b/examples/framecount.py
@@ -6,112 +6,37 @@ At logging.INFO the primary output is a string:
 "Frame # rendered at #"
 """
 
-from __future__ import division
-# float division is the expected.
-
 import logging
-# logging is primarily to set the logging level. Can also be inserted into
-# the sample model and view.
-# time is used by the view to print the unix time stamp.
-
-import ppb.engine as engine
-from ppb.event import Tick, Quit
-from ppb.utilities import Publisher
-# import ppb.engine is the primary import.
-# ppb.event to grab any events you plan to listen to.
-# ppb.utilities has multiple base classes used throughout the library.
-# in this case we'll be using a Publisher as a scene.
+import time
+import ppb
 
 
-logging.basicConfig(level=logging.INFO)
-# Set the logging level
-
-
-class TestController(object):
-    """
-    A controller that raises a Quit event after the given time.
-    """
-
-    def __init__(self, scene, length):
-        """
-        Pass in the scene and the number of seconds to run the
-        script.
-
-        :param scene: Publisher
-        :param length: number
-        :return: TestController
-        """
-        self.countdown = length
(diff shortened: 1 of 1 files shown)


astronouth7303:
This depends on #156 etc.

examples/framecount.py
```diff
@@ -6,112 +6,37 @@
...
-        :param event: Tick
-        :return:
-        """
-        self.countdown += -1 * event.sec
-        # Advance countdown. Use -1 * time to be compatible with Python2
+    def on_update(self, event, signal):
+        # Fires at the update rate (~60 times a second)
+        t = time.monotonic() - self.start_time
+        if t >= self.duration:
+            signal(ppb.events.Quit())
```
pathunstrom: Clever way to kill it.

Might be better to accumulate via Update.time_delta? (Only because it explains that feature in the sample.)

pathunstrom: Not necessary, but I wonder if it'd be worth printing both update frame rate and rendering framerate?

astronouth7303: No, because then we'd only update the time every {{tick rate}} instead of every {{frame rate}}.

astronouth7303: Possibly? If this is meant to produce useful output, interleaving them sounds like a bad idea?

astronouth7303: And I think `targets.py` adequately demonstrates Update.

pathunstrom: Um, confused at how that changes this bit of code? Using monotonic versus accumulating update time to decide when to kill the process doesn't change the print rate.

astronouth7303: Oh, I see.

You're not wrong, but there's also arguments to be made about consistency or something.

examples/framecount.py
```diff
@@ -6,112 +6,37 @@
...
-            self.count += 1
-            # Advance frame count.
-            print("Frame {} rendered at {}".format(self.count, event.run_time))
-            # Render information
+    def on_pre_render(self, event, signal):
+        # Fires each frame, which varies from the update rate
+        # We minimize the amount of work done here to maximize framerate
+        t = time.monotonic() - self.start_time
+        self.frames += 1
+        print(f"Frame {self.frames} rendered at {t}")
```
pathunstrom: So the docstring in this file mentioned log level, we should probably be using that instead of print here.

astronouth7303: That's not bad for this line. I would argue with that if we had a summary print at the end.

pathunstrom: Mostly, I want the opening docstring to match whatever the behavior is.

astronouth7303: Oh, fair.

astronouth7303:
@pathunstrom this is ready for final review.

PR #161, opened 2018-10-06
Mouse Buttons

### LINKED ISSUE #131: Mouse Button Events
Raise an event of some kind when a mouse button has been pressed but the cursor has not moved.

astronouth7303:
Implements mouse buttons.

I would like feedback as to how you feel about this style before I adjust `MouseMotion` to match.

Fixes #131.


ppb/events.py
```diff
@@ -5,6 +5,7 @@
 
 from ppb.abc import Scene
 from ppb.vector import Vector
+from .buttons import MouseButton
```
pathunstrom: Can we use absolute imports, please?

ppb/systems.py
```diff
@@ -55,9 +63,11 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     def on_update(self, update, signal):
         for pygame_event in pygame.event.get():
-            ppb_event = self.event_map.get(pygame_event.type)
-            if ppb_event is not None:
-                signal(getattr(self, ppb_event)(pygame_event, update.scene))
+            methname = self.event_map.get(pygame_event.type)
+            if methname is not None:
+                ppbevent = getattr(self, methname)(pygame_event, update.scene)
+                if ppbevent:
```
pathunstrom: This should never happen. If it's in the method map, it's on the class, and if it's not, that's on us.

ppb/systems.py
```diff
@@ -70,48 +80,27 @@ def mouse_motion(self, event, scene):
...
+    def button_pressed(self, event, scene):
+        screen_position = Vector(*event.pos)
+        camera = scene.main_camera
+        game_position = camera.translate_to_frame(screen_position)
+        btn = self.button_map.get(event.button)
+        if btn is not None:
+            return events.ButtonPressed(
+                button=btn,
+                position=game_position,
+            )
```
pathunstrom: We definitely need screen space on these.

ppb/systems.py
astronouth7303: That's not what this tests. This tests if the pygame event actually produced a ppb event.

ppb/systems.py
astronouth7303: 1. Why?
2. This isn't available on pyglet

PR #164, opened 2018-10-06
Apply magic so flags are instances

astronouth7303:
Convert flags into actual instances without sacrificing docstrings or easy declaration.

### DIFF
--- a/ppb/flags.py
+++ b/ppb/flags.py
@@ -10,28 +10,26 @@ __all__ = 'Flag', 'DoNotRender'
 
 
 class FlagMeta(type):
-    _instance = None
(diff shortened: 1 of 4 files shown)


astronouth7303:
@pathunstrom any comments on this?

ppb/flags.py
```diff
@@ -1,19 +1,58 @@
...
+        "This is a flag to indicate a thing."
+
+
+New classes of flags (eg mouse buttons) can be defined as:
+
+    class MyFlagType(Flag, abstract=True):
+        "A group of indicators"
+
+    class MyFlag(MyFlagType):
+        "This is a flag to indicate a thing."
```
pathunstrom: I like it.

PR #169, opened 2018-10-21
[nomerge] Pyglet Expirement

astronouth7303:
Tracking my pyglet experiment. This branch explicitly breaks some conventions and should be refactored before being merged.

Current problems:
* [ ] Y is flipped
* [ ] Change scanning does not detect class-level changes
* [ ] Crashes if the image doesn't exist
* [ ] `Update` is consistently late
* [ ] Breaks tests because X11 is required
* [ ] Sprite jumping if moved during PreRender (Potentially related to #170?)

Benefits:
* Zip safe
* Piles of services
* ~6x framerate over current pygame implementation


astronouth7303:
Closing this because Pyglet has blocking bugs.

PR #171, opened 2018-10-25
Fixes #168

### LINKED ISSUE #168: Non-replacing Scenes can't be popped
Something like:

```
def on_event(self, event, foo):
    event.scene.next = EphemeralScene
```

causes `EphemeralScene` to be started every time the current scene is returned to.

pathunstrom:


ppb/systems.py
```diff
@@ -118,8 +118,30 @@ class Quitter(System):
...
+        self.count = count
+
     def activate(self, engine):
-        yield events.Quit()
+        self.counter += 1
+        if self.counter >= self.count:
+            yield events.Quit()
+
+
+class FailerSystem(System):
```
astronouth7303: Why is this here and also in tests?

pathunstrom: Mistake, half a refactor.

PR #172, opened 2018-11-02
Add Keycodes

astronouth7303:
Keycodes are based on the keycode lists in pygame and pyglet.

* Letters are just the letter
* Numbers are the word
* Everything else is the full name
* Windows and Mac modifier keys were folded semantically into the the same keys
* Shifted symbols were excluded (Users should look for the Shift + base key)
* Unfortunately, only the American layout was consulted for which symbols are "base"

### DIFF
--- a/ppb/keycodes.py
+++ b/ppb/keycodes.py
@@ -1,0 +1,441 @@
+import ppb.flags
+
+
+class KeyCode(ppb.flags.Flag, abstract=True):
+    """
+    A raw keyboard scan code.
+    """
+
+
+class A(KeyCode):
+    ""
+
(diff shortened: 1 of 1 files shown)


pathunstrom:
So, one thought outside the approval: I want to figure out how to do "combined" flags in a reasonable way. I'd love an alias for `Command` that is `SuperLeft` and `SuperRight` for example.

astronouth7303:
That's... gonna take some thought? The obvious way to keep modifiers is in a `set`, eg `SuperLeft in event.modifier`, which means that multi-key constants are hard.

PR #177, opened 2018-11-16
[WIP] Interactive Shell

AstraLuma:
Add a REPL mode, allowing you to interact with a live ppb engine.

TODO:
* [ ] Ctrl+C needs to interrupt the REPL, not the engine
* [ ] Better way to edit sprite/scene classes after the fact (including updating live objects)
* [ ] Do we want the repl itself to receive events?
* [ ] Interacting with the scene is awkward. Magic `scene` builtin that gets updated?
* [ ] Can we have `print()` be interjected somehow? Updating `sys.stdout` and `sys.stderr` is fraught and tricky (since `input()`, `readline`, etc use these as well)
* [ ] Do we want to use eg ipython if its available?

### DIFF
--- a/ppb/shell.py
+++ b/ppb/shell.py
@@ -21,6 +21,7 @@ class ReplThread(threading.Thread):
             "BaseScene": ppb.BaseScene,
             "BaseSprite": ppb.BaseSprite, 
             "current_scene": self.get_scene,
+            "signal": self.signal,
         }
 
     def signal(self, event):
--- a/ppb/shell.py
+++ b/ppb/shell.py
@@ -3,6 +3,30 @@ import code
 import ppb
 import threading
 import logging
+import readline
+import pathlib
+
+
+class ReadlineHandler(logging.StreamHandler):
+    def emit(self, record):
+        # Blatently stolen from logging.StreamHandler.emit()
+
+        # ANSI for "move to the begining of the line, then erase it"
+        self.stream.write("\r\x1B[2K")
+        
+        try:
+            msg = self.format(record)
+            self.stream.write(msg + self.terminator)
+            self.flush()
+
+        except Exception:
+            self.handleError(record)
+
+        readline.redisplay()
+
+
+class BaseSprite(ppb.BaseSprite):
+    resource_path = pathlib.Path.cwd()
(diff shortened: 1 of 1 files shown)


astronouth7303:
I think the busy loop of the pygame-based engine is making this slow to react.

pathunstrom:
Haven't looked at the code yet, answering the questions I can first.

> Do we want the repl itself to receive events?

This sounds like an ouroboros to me. Do you have an idea of what we'd gain from such? (Well, aside from making a magic builtin work more easily.)

> Interacting with the scene is awkward. Magic scene builtin that gets updated?

Go for it. If we operate on the assumption that only one engine exists at a time. (We can make that assumption.)

> Can we have print() be interjected somehow? Updating sys.stdout and sys.stderr is fraught and tricky (since input(), readline, etc use these as well)

I am unsure I understand the intent of the question?

> Do we want to use eg ipython if its available?

If we can do it, I'd say this is totally a good idea.

astronouth7303:
Regards to `print()`, by "interjecting", mean "print and then redisplay the input line", so `print()` in any game objects don't disrupt the repl input.

Getting the repl to receive events isn't hard. Just write a system with an `__event__()` that looks at the repl locals instead of `self`.

astronouth7303:
So I updated this so the initial scene is tied to the repl, which is working well. These scene gets stored to the variables `scene`, and defining events handlers in the repl works perfectly.

Being able to just define event handlers is really nice. Updating them is not. Need to add a thing to allow defining multiple handlers for the same event.

Proper Ctrl-C handling feels somewhere between "black magic" and "nigh impossible", so for now I've just disabled it altogether.

Also, catch exceptions from repl-defined handlers, so they just get printed instead of crashing the entire engine.

Haven't attempted to deal with `print()` yet, but logging is handled nicely.

astronouth7303:
I'm thinking about redoing this.

First of all, I think event handlers defined in the repl should run as a system, not on some default scene, removing a lot of magic.

I could actually be convinced that running a REPL in parallel to the engine is a bad idea. This would also remove a _lot_ of magic around maintaining REPL coherency.

PR #182, opened 2018-11-16
Fixes errors related to the viewport to frame relationship.

pathunstrom:


astronouth7303:
... That's a weird one-off failure.

Issue #155, opened 2018-10-04
Set default pixel ratio to 64.


pathunstrom:
Based on ppb-mutant, and in reality, the standard size for pixel art, we should set the default pixel ratio 64.

PR #184, opened 2018-11-16
Changes pixel ration default to 64.

pathunstrom:
Completes #155

### DIFF
--- a/tests/test_camera.py
+++ b/tests/test_camera.py
@@ -39,7 +39,9 @@ def test_camera_translate_to_viewport():
 
 
 def test_sprite_in_viewport():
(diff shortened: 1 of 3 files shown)


tests/test_camera.py
```diff
@@ -39,7 +39,9 @@ def test_camera_translate_to_viewport():
 
 
 def test_sprite_in_viewport():
-    cam = Camera(viewport=(0, 0, 800, 600))
+    # Added the expected pixel ratio due to change in default breaking this test.
+    # 80 is the legacy value.
+    cam = Camera(viewport=(0, 0, 800, 600), pixel_ratio=80)
```
astronouth7303: Making sure this works off-value is a good thing.

Issue #113, opened 2018-09-13
Building a better run function


pathunstrom:
`ppb.run` has one goal: be the _easiest_ way to produce a single scene game. I'd like to fix the API instead of making it an extension of GameEngine's signature.

With that in mind, what things are critical, what things are nice to have, and what shouldn't exist?

In my mind, the `setup` function that accepts a `Scene` is the most important thing, since that's where the setup happens.

Resolution is really a nice to have, but I feel like customization of the engine is out of scope for `run`.

Additionally, do we want to add a scene that sets up invisible walls for the end user? I feel like this is a thing to leave to the end user.

pathunstrom:
Complete with https://github.com/ppb/pursuedpybear/commit/83e2365aef072fbd75f0f49131010a9a3cc41b9d

PR #186, opened 2018-11-17
Changes the signature of ppb.run to be (setup, log_level)

pathunstrom:
References #113 

Went with the thoughts mentioned in that ticket: Setup function and logging, if you want to mess with the rest of the engine, got to instantiate your own.

### DIFF
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -1,13 +1,31 @@
+import logging
+from typing import Callable
+
 from ppb.vector import Vector
 from ppb.engine import GameEngine
-
 from ppb.scenes import BaseScene
 from ppb.sprites import BaseSprite
 
-import logging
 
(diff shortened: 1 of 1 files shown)


ppb/__init__.py
```diff
@@ -1,13 +1,31 @@
...
 from ppb.vector import Vector
 from ppb.engine import GameEngine
-
 from ppb.scenes import BaseScene
 from ppb.sprites import BaseSprite
 
-import logging
 
-def run(starting_scene=BaseScene, *, log_level=logging.WARNING, **kwargs):
+def run(setup: Callable[[BaseScene], None], *, log_level=logging.WARNING):
```
astronouth7303: I actually disagree with the exclusion of the scene class/object from this signature. But I also tend to override scenes a lot.

Still, many of my demos have a scene class with event handlers of some sort.

Issue #144, opened 2018-10-03
Change Scene


pathunstrom:
The current model is. . . a bit finicky and prone to mistakes. Should probably just be via signaling.

pathunstrom:
Completed with b6dd78065a50ee70082abb5815d82607fa3825e5

PR #187, opened 2018-11-18
Change scene signals

pathunstrom:
In regards to #144

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -134,7 +134,6 @@ class GameEngine(Engine, EventMixin, LoggingMixin):
         """
         Start a new scene. The current scene pauses.
         """
-        # Empty the queue to limit the amount that can happen after a change.
         self.pause_scene()
         self.start_scene(event.new_scene, event.kwargs)
 
@@ -160,13 +159,13 @@ class GameEngine(Engine, EventMixin, LoggingMixin):
 
     def pause_scene(self):
         # Empty the queue before changing scenes.
-        self.events = deque()
+        self.flush_events()
         self.signal(events.ScenePaused())
         self.publish()
 
     def stop_scene(self):
         # Empty the queue before changing scenes.
-        self.events = deque()
+        self.flush_events()
         self.signal(events.SceneStopped())
         self.publish()
         self.scenes.pop()
@@ -179,3 +178,12 @@ class GameEngine(Engine, EventMixin, LoggingMixin):
 
     def register(self, event_type, attribute, value):
         self.event_extensions[event_type][attribute] = value
+
+    def flush_events(self):
+        """
+        Flush the event queue.
+
+        Call before doing anything that will cause signals to be delivered to
+        the wrong scene.
+        """
+        self.events = deque()
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -12,8 +12,6 @@ from ppb.abc import Engine
 from ppb.events import ChangeScene
 from ppb.events import EventMixin
 from ppb.events import Quit
-from ppb.events import SceneStart
-from ppb.events import ScenePause
 from ppb.systems import PygameEventPoller
 from ppb.systems import Renderer
 from ppb.systems import Updater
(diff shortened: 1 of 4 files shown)


ppb/engine.py
```diff
@@ -120,8 +130,52 @@ def manage_scene(self):
...
+        """
+        self.stop_scene()
+        self.start_scene(event.new_scene, event.kwargs)
+
+    def on_quit(self, quit_event: Quit, signal: Callable[[Any], None]):
         self.running = False
 
+    def pause_scene(self):
+        # Empty the queue before changing scenes.
+        self.events = deque()
```
astronouth7303: Can we make this a method? It feels deserving of its own method.

ppb/engine.py
```diff
@@ -120,8 +130,52 @@ def manage_scene(self):
         if next_scene:
             self.activate(next_scene)
 
-    def on_quit(self, quit_event: 'Quit', signal: Callable):  #TODO: Look up syntax for Callable typing.
+    def on_change_scene(self, event: ChangeScene, signal: Callable[[Any], None]):
+        """
+        Start a new scene. The current scene pauses.
+        """
+        # Empty the queue to limit the amount that can happen after a change.
```
astronouth7303: Stale comment

ppb/events.py
```diff
@@ -1,19 +1,30 @@
...
 from typing import Set
+from typing import Type
+from typing import Union
 
 from ppb.abc import Scene
 from ppb.buttons import MouseButton
 from ppb.keycodes import KeyCode
 from ppb.vector import Vector
 
 __all__ = (
```
astronouth7303: FYI, I don't feel a particular need to keep this alphabetical. Thematic sorting is just as reasonable.

ppb/events.py
```diff
@@ -98,6 +109,23 @@ class ButtonReleased:
     scene: Scene = None
 
 
+@dataclass
+class ChangeScene:
+    """
+    Fired to start a new scene.
+
+    new_scene can be an instance or a class. If a class, must include kwargs.
+    If an instance kwargs must be falsey, default is None.
```
astronouth7303: Eh?

ppb/events.py
```diff
@@ -98,6 +109,23 @@ class ButtonReleased:
...
+
+    Examples:
+        * `signal(new_scene=ChangeScene(MyScene(player=player))`
+        * `signal(new_scene=ChanngeScene, kwargs={"player": player}`
```
astronouth7303: I kinda wish this form had a calling form, like `SomethingSomething(ChangeScene, player=player)`.

Also, typo

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+class ReplaceScene:
+    """
+    Fired to replace the current scene with a new one.
+
+    new_scene can be an instance or a class. If a class, must include kwargs.
+    If an instance kwargs must be falsey, default is None.
+
+    Examples:
+        * `signal(new_scene=ChangeScene(MyScene(player=player))`
+        * `signal(new_scene=ChanngeScene, kwargs={"player": player}`
```
astronouth7303: Missed copy/paste replacements

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+    """
```
astronouth7303: Should probably mention `SceneStopped`

ppb/events.py
```diff
@@ -98,6 +109,23 @@ class ButtonReleased:
...
+    Fired to start a new scene.
+
+    new_scene can be an instance or a class. If a class, must include kwargs.
+    If an instance kwargs must be falsey, default is None.
+
+    Examples:
+        * `signal(new_scene=ChangeScene(MyScene(player=player))`
+        * `signal(new_scene=ChanngeScene, kwargs={"player": player}`
+    """
```
astronouth7303: Should probably mention `ScenePaused`

ppb/events.py
```diff
@@ -98,6 +109,23 @@ class ButtonReleased:
     scene: Scene = None
 
 
+@dataclass
+class ChangeScene:
```
astronouth7303: I feel like this is a insufficiently distinct name from `ReplaceScene`. Either make it more clear this is adding to the stack or explicitly document it?

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+    Fired when a scene starts.
+    """
+    scene: Scene = None
+
+
+@dataclass
+class SceneStopped:
+    """
+    Fired when a scene stops.
+    """
```
astronouth7303: Make it more clear that this is fired when a scene dies, so each scene will only receive this once?

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+    scene: Scene = None
+
+
+@dataclass
+class ScenePaused:
+    """
+    Fired when a scene pauses.
+    """
```
astronouth7303: Should mention `SceneContinued`

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+    new_scene: Union[Scene, Type[Scene]]
+    kwargs: Dict[str, Any] = None
+    scene: Scene = None
+
+
+@dataclass
+class SceneContinued:
+    """
+    Fired when a paused scene continues.
+    """
```
astronouth7303: Should probably mention `ScenePaused`

ppb/events.py
astronouth7303: Should probably also mention `SceneStarted`

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+    scene: Scene = None
+
+
+@dataclass
+class SceneStarted:
+    """
+    Fired when a scene starts.
+    """
```
astronouth7303: Should probably also mention `SceneStopped`.

ppb/events.py
astronouth7303: Oh, and when this event fires, the stopped scene is still the current scene.

ppb/events.py
```diff
@@ -145,6 +174,63 @@ class Render:
...
+    Fired when a scene pauses.
+    """
+    scene: Scene = None
+
+
+@dataclass
+class StopScene:
+    """
+    Fired to stop a scene.
+    """
```
astronouth7303: I think the above comments make it clear that more documentation around the scene life cycle events is needed.

tests/test_engine.py
```diff
@@ -68,7 +69,7 @@ def test_next_scene_none(self):
         self.assertIs(self.engine.current_scene, self.mock_scene)
 
 
-def test_scene_change():
+def test_scene_change_thrashing():
```
astronouth7303: Should this test be updated for the event-based way of doing things? Or maybe copied/updated?

ppb/events.py
pathunstrom: We could make it _not_ a dataclass? Then I can just add kwargs to that signature.

It makes it feel weird because it'll only be a handful of events that do that.

ppb/events.py
pathunstrom: First one I named. We can just make it `StartScene`?

tests/test_engine.py
pathunstrom: This shouldn't affect signal based scene changes, since it requires an action from the user instead of an implicit result of setting values. So because we won't be using `Sprite.next` to do anything (once the old code is removed) it's not a thing that should happen.

pathunstrom:
So, there is a single test testing each of the ways you can switch scenes with the old method. I know because they were added because the API was reversed at one point. (`running=False` kept the old scene running).

I don't think we need test parity since they happen using completely different code and are separate APIs.

There is one feeling I'm having regarding events: Right now, it's hard to type hint certain things due to circular imports due to the EventMixin. Also, because we're developing two distinct kinds of events, we have a logical break between them that could be separated at the module level. It might be time to pull all of these apart into multiple modules.

ppb/events.py
pathunstrom: This better?

astronouth7303: It's still the wrong name.

PR #195, opened 2018-12-17
[WIP] Initial sphinx docs

astronouth7303:
An initial run at Sphinx and ReadTheDocs.


pathunstrom:
@astronouth7303 Got a big question: When does readthedocs do a new build? Is it "per release" or every update to master?

astronouth7303:
Depends on the rtd configuration, because rtd has rich version/tag support.

In general, I think it's rebuilt if a given branch or tag is updated.

astronouth7303:
Also, I feel like a lot more needs to be written? Unless you're ok with this as a starting point.

PR #204, opened 2019-01-13
Immutable vectors

### LINKED ISSUE #188: Make Vectors immutable
Should we just make vectors immutable things? That seems like best practice?

astronouth7303:
Refactor the sides APIs to not mutate Vectors anymore.

Closes #188 

(Note: This is WIP because it depends on Vector 1.0 APIs. It is otherwise ready for review.)


ppb/sprites.py
```diff
@@ -127,7 +127,7 @@ def center(self, value):
         self.parent.position = self._mk_update_vector_center(value)
 
     def _mk_update_vector_side(self, attribute, value):
-        value = Vector(*value)
+        value = Vector.convert(value)
```
nbraud: `Vector(value)` or `Vector(*value)` should work perfectly fine, no?

astronouth7303: This was written before the copy constructor was merged.

nbraud: 'k.  To be fair, `Vector.convert()` might be avoiding a copy when `value` is already a `Vector`.

astronouth7303: By design, yes.

nbraud: Yes, I was just saying that I just realised using `Vector.convert` is better there.
Would it be possible to avoid the copy when using `Vector()` ?

astronouth7303: It's conventional for the `MyType(value)` to be a copy constructor, like `list()`, `dict()`, `set()`, etc

astronouth7303: So, this is conventional for mutable types.

It looks like immutable types tend to not copy.

ppb/sprites.py
```diff
@@ -173,7 +200,7 @@ def __init__(self, **kwargs):
                 k = 'position'
             # Castings
             if k in ('position', 'facing'):
-                v = Vector(*v)  # Vector.convert() when that ships.
+                v = Vector.convert(v)
```
nbraud:
```suggestion
                v = Vector(v)
```

ppb/sprites.py
```diff
@@ -125,12 +124,40 @@ def center(self):
...
     def center(self, value):
-        if self.side in (TOP, BOTTOM):
-            setattr(self.parent, self.side, value[1])
-            self.parent.center.x = value[0]
-        else:
-            setattr(self.parent, self.side, value[0])
-            self.parent.position.y = value[1]
+        self.parent.position = self._mk_update_vector_center(value)
+
+    def _mk_update_vector_side(self, attribute, value):
```
nbraud: Add a docstring documenting what this does?  It's kinda unobvious from the name or the code.

ppb/sprites.py
```diff
@@ -125,12 +124,40 @@ def center(self):
...
+
+        assert self_dimension != attr_dimension
+
+        fields = {
+            self_dimension: value[self_dimension] - self_offset,
+            attr_dimension: value[attr_dimension] - attr_offset,
+        }
+        return Vector(**fields)
+
+    def _mk_update_vector_center(self, value):
```
nbraud: Same, add docstring.

ppb/sprites.py
```diff
@@ -185,49 +212,46 @@ def center(self) -> Vector:
 
     @center.setter
     def center(self, value: Sequence[float]):
-        x = value[0]
-        y = value[1]
-        self.position.x = x
-        self.position.y = y
+        self.position = Vector.convert(value)
```
nbraud:
```suggestion
        self.position = Vector(value)
```

nbraud:
@astronouth7303 Would it be possible to make it work with both versions of `ppb-vector`?

Currently the vector 1.0 release should block on this (because `pursuedpybear` doesn't specify a version constraint, so users and CI might upgrade at any point after the release), but this can't be merged before we cut a new version.

nbraud:
PS: Once #232 lands, this branch should update the version dependency on `ppb-vector` (probably to `~= 1.0`)

ppb/sprites.py
```diff
@@ -125,12 +124,40 @@ def center(self):
...
+
+        attr_dimension, attr_offset = self._lookup_side(attribute)
+
+        assert self_dimension != attr_dimension
+
+        fields = {
+            self_dimension: value[self_dimension] - self_offset,
+            attr_dimension: value[attr_dimension] - attr_offset,
+        }
+        return Vector(**fields)
```
nbraud: Is this correct?  I don't think `Vector()` would accept keyword arguments `attr_dimension` and `self_dimension`

ppb/sprites.py
```diff
@@ -185,49 +212,46 @@ def center(self) -> Vector:
 
     @center.setter
     def center(self, value: Sequence[float]):
```
nbraud:
```suggestion
    def center(self, value: ppb_vector.vector2.VectorLike):
```

This change should likely wait until we cut a release of `ppb-vector` though.

ppb/sprites.py
astronouth7303: no, they're equal to `'x'` and `'y'`.

ppb/sprites.py
```diff
@@ -209,43 +246,40 @@ def center(self) -> Vector:
         return self.position
 
     @center.setter
-    def center(self, value: Sequence[float]):
-        x = value[0]
-        y = value[1]
-        self.position.x = x
-        self.position.y = y
+    def center(self, value: ppb_vector.vector2.VectorLike):
```
nbraud:
```
AttributeError: module 'ppb_vector.vector2' has no attribute 'VectorLike'
```
Looks like we need to expose `VectorLike` in `ppb-vector` explicitely?

astronouth7303: That's not how Python namespaces work.

```
pip list
Package        Version
-------------- -------
atomicwrites   1.3.0  
attrs          19.1.0 
cffi           1.12.2 
dataclasses    0.6    
greenlet       0.4.13 
more-itertools 7.0.0  
pip            19.0.3 
pluggy         0.9.0  
ppb-vector     0.3    
py             1.8.0  
pygame         1.9.5  
pytest         4.4.0  
readline       6.2.4.1
setuptools     40.8.0 
six            1.12.0 
wheel          0.33.1 
```

nbraud: Oh, right, that's the bug with Cirrus being somehow worse at caching than Docker itself.
*needs to report the bug, still.*

astronouth7303:
I would like to get this merged in. Maybe we can rollback the docker stuff until Cirrus improves their container caching?

pathunstrom:
I'm okay with that. @nbraud That was your work, what's your thoughts?

astronouth7303:
bors retry

PR #209, opened 2019-02-06
updated repr function to new format style

dillonko:
In order to help with the uniformity of the file I updated the __repr__ function. If there is an other reason that repr was not changed please disregard. Otherwise, Thank you for your consideration.


dillonko:
Ok thanks will do!

astronouth7303:
Also, if you'd like, add yourself as a Contributor to CONTRIBUTORS.md

Issue #167, opened 2018-10-14
Resource path extension breaks


pathunstrom:
If you override resource_path you get the following error, stack trace anonymized:

```
Traceback (most recent call last):
  File "scratch.py", line 11, in <module>
    run(scene_kwargs={"set_up": setup})
  File ".venv/lib/python3.7/site-packages/ppb/__init__.py", line 13, in run
    eng.run()
  File ".venv/lib/python3.7/site-packages/ppb/engine.py", line 78, in run
    self.main_loop()
  File ".venv/lib/python3.7/site-packages/ppb/engine.py", line 92, in main_loop
    self.publish()
  File ".venv/lib/python3.7/site-packages/ppb/engine.py", line 113, in publish
    entity.__event__(event, self.signal)
  File ".venv/lib/python3.7/site-packages/ppb/events.py", line 62, in __event__
    meth(bag, fire_event)
  File ".venv/lib/python3.7/site-packages/ppb/systems.py", line 149, in on_render
    resource = self.prepare_resource(game_object)
  File ".venv/lib/python3.7/site-packages/ppb/systems.py", line 164, in prepare_resource
    self.register_renderable(game_object)
  File "/.venv/lib/python3.7/site-packages/ppb/systems.py", line 191, in register_renderable
    self.register(source_path / image_name, image_name)
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

Here's the minimum sample to reproduce:

```python
from ppb import BaseSprite, run


class Player(BaseSprite):
    resource_path = "./resources/"


def setup(scene):
    scene.add(Player())

run(scene_kwargs={"set_up": setup})
```

```
$ python --version
Python 3.7.0
```

dillonko:
I get a different error:

```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "game_issue.py", line 9, in <module>
    run(scene_kwargs={"set_up": setup})
TypeError: run() got an unexpected keyword argument 'scene_kwargs'
```

```
$ python --version
Python 3.7.1
```

astronouth7303:
I think this is part of the bigger problem that we want typed properties (see also: assigning tuples to position)

pathunstrom:
This one is more on the protocol side. We want it to take a string, but it needs to be transformed to a Path in the protocol to work the way we intend.

dillonko:
Can you give me more insight on that. It will be easier to execute if I
[...]

PR #211, opened 2019-02-21
This update appears to resolve issue #167

dillonko:
RESOLVES: Resource path extension breaks #167. Please let me know if there are any other areas I might be overlooking.


dillonko:
where is the path to images supposed to be? I will probably have a better solution tonight.

pathunstrom:
So, you've got 4 pieces:

Every sprite has an `image` and `resource_path` attribute.

This attributes are the default values to the `__image__` and __resource_path__` protocols.

My guess is the best place to fix this bug is to make sure `__resource_path__` returns a pathlib.Path object.

PR #213, opened 2019-03-15
ppb/vector: Define __slots__

nbraud:
When used with a version of ppb-vector that defines `__slots__`, this dramatically reduces memory usage.

When used with an earlier version of ppb-vector, `__dict__` is accessible, making the change a no-op.

See https://github.com/ppb/ppb-vector/pull/115

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,3 +19,4 @@ Name | Email | Twitter |
 [Dimitris Maroulidis](https://github.com/dmaroulidis) | [dimitris@dimitrismaroulidis.com](mailto:dimitris@dimitrismaroulidis.com) | [dmaroulidis](https://twitter.com/dmaroulidis)
 [Jenny Li](https://github.com/imjennyli) | 
 [Andrew Schamp](https://github.com/schamp/) | | [schamp](https://twitter.com/schamp)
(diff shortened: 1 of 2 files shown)


astronouth7303:
This is blocked on a release of ppb-vector.

nbraud:
@astronouth7303 I don't think this should block on a `ppb-vector` release; as said above, it's just a no-op on current versions of `ppb-vector`.

If this broke with previous versions of `ppb-vector`, it would need an update of the dependencies metadata (and would likely break for ppb users due to the `pip` bug we keep running into).

astronouth7303:
Nope, just ran a test:

```
>>> class A: pass
>>> class B:
...   __slots__ = ()
>>> b = B()
>>> b.x = 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'B' object has no attribute 'x'
>>> vars(b)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: vars() argument must have __dict__ attribute
```

nbraud:
Because you aren't making `B` inherit from an unslotted class:
```
>>> class Vector2: pass

>>> class Vector(Vector2):
        __slots__ = ()

>>> Vector().x = 0
```

I did actually test it before saying it works...  :P

astronouth7303:
oh, oops, i misdid my demo.

PR #214, opened 2019-03-25
Add rotatability to sprites.

### LINKED ISSUE #142: Sprite Rotation
We should rotate sprites to their facing before rendering!

In regards to #141: Scale before rotation.

pathunstrom:
First pass at this. Will still need to add the functionality to rotate the images themselves.

Closes #142

### DIFF
--- a/ppb/sprites.py
+++ b/ppb/sprites.py
@@ -163,7 +163,7 @@ class Rotatable:
         self._rotation = value % 360
 
     def rotate(self, degrees):
-        """Change the current rotation by degrees."""
+        """Rotate the sprite by a given angle (in degrees)."""
         self.rotation += degrees
 
 
--- a/ppb/sprites.py
+++ b/ppb/sprites.py
@@ -146,11 +146,13 @@ class Rotatable:
     """
     _rotation = 0
     # This is necessary to make facing do the thing while also being adjustable.
-    _basis = Vector(0, -1)
+    basis = Vector(0, -1)
+    # Considered making basis private, the only reason to do so is to
+    # discourage people from relying on it as data.
 
     @property
     def facing(self):
-        return Vector(*self._basis).rotate(self.rotation).normalize()
+        return Vector(*self.basis).rotate(self.rotation).normalize()
 
     @property
     def rotation(self):
--- a/ppb/sprites.py
+++ b/ppb/sprites.py
@@ -182,7 +182,6 @@ class BaseSprite(EventMixin, Rotatable):
     image = None
     resource_path = None
     position: Vector = Vector(0, 0)
-    facing: Vector = Vector(0, -1)
     size: Union[int, float] = 1
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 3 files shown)


ppb/sprites.py
```diff
@@ -140,7 +140,32 @@ def _attribute_gate(self, attribute, bad_sides):
...
 
 
-class BaseSprite(EventMixin):
+class Rotatable:
+    """
+    A simple rotation mixin. Can be included with sprites.
+    """
+    _rotation = 0
+    # This is necessary to make facing do the thing while also being adjustable.
+    _basis = Vector(0, -1)
```
astronouth7303: Is this meant to be public?

pathunstrom:
Assuming it passes Travis, this should be ready to go.

ppb/sprites.py
```diff
@@ -140,7 +140,34 @@ def _attribute_gate(self, attribute, bad_sides):
...
+    def facing(self):
+        return Vector(*self.basis).rotate(self.rotation).normalize()
+
+    @property
+    def rotation(self):
+        return self._rotation
+
+    @rotation.setter
+    def rotation(self, value):
+        self._rotation = value % 360
```
nbraud: This can still be negative; should this use the same convention as `ppb-vector` and have `-180 < _rotation <= 180`?

ppb/sprites.py
```diff
@@ -140,7 +140,34 @@ def _attribute_gate(self, attribute, bad_sides):
...
+
+    def rotate(self, degrees):
+        """Change the current rotation by degrees."""
```
nbraud:
```suggestion
        """Rotate the sprite by a given angle (in degrees)."""
```

ppb/sprites.py
```diff
@@ -140,7 +140,34 @@ def _attribute_gate(self, attribute, bad_sides):
             raise AttributeError(message)
 
 
-class BaseSprite(EventMixin):
+class Rotatable:
```
nbraud: `Rotatable` sounds weird to me, but I'm a silly non-native speaker.  *shrugs*

pathunstrom: Native speaker, sounds weird to me, too, but lots of mixins are named based on their capabilities.

ppb/sprites.py
pathunstrom: In the case of this interface, dunno if it matters. The tests have some negative rotation and it gives the equivalent positive rotation, which is a reasonable result, but if both you and @astronouth7303 feel like I should port that over here, I will.

nbraud: @pathunstrom I don't think it's much of an issue either way, I just wanted to check that was intentional.

ppb/sprites.py
astronouth7303: -able is approximately the right convention, but it leads to lots of weird words.

I would also accept CanBe*? Also weird.

pathunstrom: So like `CanBeRotated`?

nbraud: Or just `CanRotate`?  In any case, I don't think `Rotatable` is a problem, I was just very confused whether there was a type, or it was just weird, or I was just bad at English.

PR #215, opened 2019-03-26
Improve flakiness of Failer tests

astronouth7303:
Try to improve the flakiness of the failer timing-based tests.

### DIFF
--- a/tests/test_testutil.py
+++ b/tests/test_testutil.py
@@ -17,22 +17,28 @@ def test_quitter(loop_count):
     raise AssertionError("Quitter did not raise a quit event.")
 
 
-def test_failer():
+def test_failer_immediate():
     failer = testutil.Failer(fail=lambda e: True, message="Expected failure.")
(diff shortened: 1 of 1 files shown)


astronouth7303:
So, it's still supposed to be within 0.01s of the prescribed time, but I can't reproduce the bug.

Is this version of the test more clear?

pathunstrom:
I preferred the `with raises` but that's not really a problem either way. I like splitting the test in two, nice touch.

The test has been flaky since the beginning, we might have to run the tests a handful of times to guarantee it. (This is that one test that just. . . fails sometimes. Especially on Travis.)

astronouth7303:
I'll rerun them both a few more times, then.

astronouth7303:
Haven't been able to reproduce so far, so yay?

pathunstrom:
Yeah, let's keep this and if it comes back we can approach it again.

PR #216, opened 2019-03-27
Setup Cirrus CI

nbraud:
- [x] `pytest` on the same platforms as `ppb-vector`
- [x] Added a `docs/requirements.txt` file
- [x] Build docs in a separate task.
- [x] Install `pygame`'s build dependencies
- [x] Only install build dependencies if no binary wheel is available.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,7 +4,7 @@ docs_task:
     image: python:3.7-slim
 
   install_script:
-    - apt update && apt install make
+    - apt update || true ; apt install -qq -y make
     - pip install --upgrade-strategy eager -U -r requirements-docs.txt
     - pip install -e .
 
@@ -46,6 +46,8 @@ Linux_task:
       - image: pypy:3.6-slim
 
   install_script:
+    - apt update || true
+    - apt install -qq -y pkgconf libsdl-image1.2-dev libsdl-mixer1.2-dev libsdl-ttf2.0-dev libsdl1.2-dev libsmpeg-dev libportmidi-dev libswscale-dev libavformat-dev libavcodec-dev libfreetype6-dev gcc
     - pip install --upgrade-strategy eager -U pytest
     - pip install -e .
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,29 +12,6 @@ docs_task:
     - make -C docs/ html
 
 
-FreeBSD_task:
-  skip: $CIRRUS_BRANCH =~ '.*\.tmp'
-  freebsd_instance:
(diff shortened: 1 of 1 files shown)


nbraud:
@pathunstrom I forgot to leave a comment yesterday, but this should be good-to-go; the Windows/Py3.6 failure is the timing issue that already exists on `master` (though I'd totally understand waiting for that to be fixed so you can check Cirrus goes green)

nbraud:
@astronouth7303 I just noticed there already was a `requirements-docs.txt`, so I just removed `docs/requirements.txt`.

Do you think it would be reasonable to rename that file though?  I didn't expect `requirements-*.txt` (and instead `*-requirements.txt` or `*/requirements.txt`), hence why I didn't notice it, and I have the impression the latter is more common than the former (`test-requirements`, `dev-requirements`, ...)

nbraud:
I'm experimenting with [caching dependencies] in custom container images in https://github.com/nbraud/pursuedpybear/pull/1

[caching dependencies]: https://medium.com/cirruslabs/introducing-container-builder-for-cirrus-ci-80b9234f007

nbraud:
Update: I got the test time (for Linux tasks) down to ~2s in nbraud#1, from resp. ~10s (if there's a `pygame` wheel) and 2 minutes (when there is no wheel).

I'm going to try and see if I can easily make the same thing for Windows containers.
Would you prefer that work in this PR directly, or in a follow-up PR?  (I'd prefer the latter, but it's no strong preference)

pathunstrom:
We can definitely do that as a follow on pr.

And @astronouth7303 there's that windows failure again.

pathunstrom:
@nbraud You comfortable with this? It seems to be functioning and I'd like to get it in since it'll make our lives better on the rest of the work getting done.

nbraud:
PS: I forgot to mention yesterday, but I believe nbraud#1 is ready.  I did generalise the `Dockerfile`s to Windows containers, and got the test time there down to ~5s.  I didn't yet address the issues we discussed on Discord about automatically picking up new versions of dependencies from PyPI, and I think that would involve using entirely different tooling, so that's basically it for “v1” of dependency caching.

macOS doesn't support anything like containers, so the same approach won't work there (though @astronouth7303 mentioned that it might be possible to persist the `pip` cache and perhaps even a whole virtualenv)

pathunstrom:
Is nbraud#1 supposed to be against your fork?

nbraud:
Yeah, I didn't want to add all of it to this PR right away, but I still wanted to have Cirrus run it; I'll open a new PR from that branch once this lands, I guess.

PR #220, opened 2019-03-31
WiP: Convert camera tests to Hypothesis

nicoonoclaste:
- [x] `test_camera_viewport`
- [ ] `test_camera_point_in_viewport_not_at_origin`
- [ ] `test_camera_translate_to_frame`
- [ ] `test_camera_translate_to_viewport`
- [ ] `test_sprite_in_viewport`
- [ ] `test_viewport_change_affects_frame_height`

### DIFF
--- a/ppb/testutils.py
+++ b/ppb/testutils.py
@@ -16,6 +16,13 @@ def integer_vectors(min_value=None, max_value=None):
         st.integers(min_value=min_value, max_value=max_value),
     )
 
+def vectors(max_magnitude=1e75):
+    return st.builds(
+        Vector,
+        st.floats(min_value=-max_magnitude, max_value=max_magnitude),
+        st.floats(min_value=-max_magnitude, max_value=max_magnitude),
+    )
+
 
 class Failer(System):
 
--- a/ppb/testutils.py
+++ b/ppb/testutils.py
@@ -1,9 +1,20 @@
 import time
 from typing import Callable
 
+from hypothesis import strategies as st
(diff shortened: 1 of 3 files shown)


nbraud:
I'm having some trouble converting `test_camera_point_in_viewport_not_at_origin` to an Hypothesis test.

Now I'm unsure I understand correctly what `Camera.point_in_viewport` is supposed to do (it's either that or there's a bug in it, given that I can't get Hypothesis not to explode on that test), and I can't find doc that clarifies it  :(

pathunstrom:
I think the correct answer to what it's supposed to do is "be overengineering made manifest"

It literally checks if a given vector exists in the viewport/window the camera is attached to. As far as I can tell, it's not been used and existed to handle the future case of multiple cameras.

nbraud:
@pathunstrom OK; I still don't quite get what it's supposed to do (or in which coordinate system it operates, frame or viewport) so I will leave that test alone for now.

pathunstrom:
Pretty sure it's pure viewport, not even a translation from one to the other. That's my bad.

nbraud:
FWIW, CI is failing because I needed `Vector.isclose`, which isn't in a released version of `ppb-vector` yet.

AstraLuma (commented):
Seems good so far.

astronouth7303:
Looking this over, I don't think #439 covers everything this does/aspires to be.

So I'm going to leave this open for now.

PR #221, opened 2019-04-01
New subsystem API

pathunstrom:
So, doing what I discussed in the slack: Removing activate, adding a heartbeat event.

Need to decide if we heartbeat between each publish, or we clear the queue between each heartbeat.

If we do the former, we're going to start seeing scheduling silliness, but that just comes from asynchronous design.

If we do the latter, it's easier to make events happen in a sequence, but still going to have some of the asynchronous silliness because, well, we're technically async.

- [x] Add heartbeat, switch the test utilities to use a heartbeat system.
- [x] Remove activate, switch remaining systems to use heartbeat system.
- [x] Clean up the main loop to reduce the processing time of the busy loop.

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -36,6 +36,7 @@ class GameEngine(Engine, EventMixin, LoggingMixin):
         self.event_extensions = defaultdict(dict)
         self.running = False
         self.entered = False
+        self.last_heartbeat = None
 
         # Systems
         self.systems_classes = systems
@@ -80,12 +81,20 @@ class GameEngine(Engine, EventMixin, LoggingMixin):
 
     def start(self):
         self.running = True
+        self.last_heartbeat = time.monotonic()
         self.activate({"scene_class": self.first_scene,
                        "kwargs": self.scene_kwargs})
 
     def main_loop(self):
         while self.running:
             time.sleep(0)
+            now = time.monotonic()
+            self.signal(events.Heartbeat(now - self.last_heartbeat))
+            self.last_heartbeat = now
+            # Okay, doing this twice is hacky as hell, but fine while I remove
+            # the old system.
+            while self.events:
+                self.publish()
             for system in self.systems:
                 for event in system.activate(self):
                     self.signal(event)
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -36,7 +36,7 @@ class GameEngine(Engine, EventMixin, LoggingMixin):
         self.event_extensions = defaultdict(dict)
         self.running = False
         self.entered = False
-        self.last_idle_event = None
+        self._last_idle_time = None
 
         # Systems
         self.systems_classes = systems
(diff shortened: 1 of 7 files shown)


astronouth7303:
This is meant as an Idle event? Could it be an event that's triggered not through the `signal()` path but by the event loop itself?

astronouth7303:
Idle implies "when the queue is empty"

pathunstrom:
That's the question in the beginning, is it better as an idle event or as simply a timing event in between?

It needs to be `signal`ed currently because publish does the popping of the queue.

ppb/events.py
```diff
@@ -271,6 +271,15 @@ class StopScene:
     scene: Scene = None
 
 
+@dataclass
```
nbraud:
```suggestion
@dataclass(frozen=True)
```

astronouth7303: We actually can't freeze events, because of the feature allowing systems to modify events.

nbraud: OK, ok.

pathunstrom:
I guess, as a followup, it doesn't _have_ to be the way it is now, but that's. . . at least two steps forward in the process. This work gets the heartbeat _working_ and updates Failer to use the thing. The next step would be removing activate and updating the rest of the systems. From there, we can change how publish works and move the pop into the main loop, and let it publish all before the next heartbeat.

So. . . summary:

This is the first (workable) step of a series:
1. Add heartbeat, switch the test utilities to use a heartbeat system.
2. Remove activate, switch remaining systems to use heartbeat system.
3. Clean up the main loop to reduce the processing time of the busy loop.

Theoretically, we can break it up further, but I think this is the right amount of breaking up to be able to properly test things in the interim.

pathunstrom:
I think this one is good for review.

ppb/engine.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, first_scene: Type, *,
         self.event_extensions = defaultdict(dict)
         self.running = False
         self.entered = False
+        self.last_idle_event = None
```
astronouth7303: This is poorly named.

ppb/systems/__init__.py
```diff
@@ -141,12 +152,13 @@ def __init__(self, time_step=0.016, **kwargs):
     def __enter__(self):
         self.start_time = time.time()
 
-    def activate(self, engine):
+    def on_idle(self, idle_event: events.Idle, signal):
         if self.last_tick is None:
             self.last_tick = time.time()
```
astronouth7303: Is `time.time()` the right thing?

pathunstrom: That's what it used to be. I can update that to monotonic.

ppb/engine.py
pathunstrom: It started as "last_heartbeat". Would "last_idle_timestamp" be better?

ppb/systems/__init__.py
astronouth7303: I know it's not strictly in-scope, but while you're here, could you?

PR #223, opened 2019-04-01
setup.py: Improve dependencies handling

### LINKED ISSUE #222: Add Test Dependencies to setup.py
Didn't realize I hadn't done this, we should absolutely have the setup.py set up to manage test dependencies.

nbraud:
- [x] Make dependencies available in `requirements.txt`.
  This is necessary to be able to build CI environments as in nbraud#1.
- [x] Add a list of test dependencies (in `requirements-tests.txt`).  Closes #222.
- [x] Integrate `setup.py` and `pytest` so calling `setup.py test` does the right thing.


nbraud:
I had to create a `setup.cfg` to define the necessary alias for `pytest-runner` to be called by `setup.py test`, but I didn't try to convert more of `setup.py` to `setup.cfg`, as that felt out-of-scope.

I can do so in a follow-up PR though  :)

nbraud:
PS: The file is called `requirements-tests.txt` rather than `test-requirements.txt` because there's already a `requirements-docs.txt`; I asked [in #216](https://github.com/ppb/pursuedpybear/pull/216#issuecomment-478236102) whether the convention should be changed, but got no answer.

nbraud:
BTW, once this is merged, #216 and #220 should be updated.

PR #225, opened 2019-04-02
CI: Build custom containers with dependencies preinstalled

nbraud:
Brings the CI times down to ~2s on Linux (from ~2m for envs with no `pygame` wheels), and to ~10s on Windows (from ~30s with `pygame` wheels).

- [x] Build & use containers for tests under Linux and Windows.
- [x] Build & use containers for the `docs` ~and `lint`~ tasks.
- [x] Add a CI task to check the generated dockerfiles are up-to-date.
  I'd prefer not to commit generated files at all, but that was the simplest way to get going (Cirrus can just pull a Dockerfile from the repo)
- [x] Give friendly names to the tasks using custom containers.


nbraud:
Rebased again to have `bors.toml` (necessary because 009fe2c adds a new, required check)

nbraud:
bors r=pathunstrom

PR #228, opened 2019-04-03
Fix registration

pathunstrom:
Changes GameEngine.register from accepting values to accepting callbacks. This allows dynamic values to be applied as part of event publication extension.


ppb/engine.py
```diff
@@ -33,7 +35,7 @@ def __init__(self, first_scene: Type, *,
         # Engine State
         self.scenes = []
         self.events = deque()
-        self.event_extensions = defaultdict(dict)
+        self.event_extensions: DefaultDict[type, Dict[str, Callable[[], Any]]] = defaultdict(dict)
```
nbraud: I don't think the type annotation is useful (it should be at the class level for `mypy` to be able to do something with it.)

pathunstrom: I put it in because my linter was whining about types down in the publish function.

ppb/engine.py
```diff
@@ -108,8 +110,9 @@ def publish(self):
         event = self.events.popleft()
         scene = self.current_scene
         event.scene = scene
-        for attr_name, attr_value in self.event_extensions[type(event)].items():
-            setattr(event, attr_name, attr_value)
+        extension_dict = self.event_extensions[type(event)]
+        for attr_name, attr_callback in extension_dict.items():
+            setattr(event, attr_name, attr_callback())
```
astronouth7303: Shouldn't we pass it the event that's getting the attribute?

ppb/engine.py
nbraud: OK, that's weird.

astronouth7303:
You haven't implemented the type wildcard?

pathunstrom:
Right. Will get on that.

PR #229, opened 2019-04-05
Universal Tree Model.

pathunstrom:
This isn't anywhere near done, but this is the thing I'm considering doing.

- [x] ~~Add a children property of the engine - Will deliver current scene and subsystems~~
- [x] Need to decide if we handle an event on the parent before the children or vice versa.
- [x] Pulled the children publication out of EventMixin to a subclass.
- [x] ~~Make a change to how the renderer deals with these so it can render children sprites without them being contained in the scene.~~
- [x] Add new mixin to GameEngine
- [x] Add new mixin to Scene
- [x] Test mixin
- [x] Test GameEngine with mixin
- [x] Test Scene with mixin


astronouth7303:
I also feel like the engine should have all of the scenes as its children? However, it can have `.current_scene` and `.systems` as children views.

I'm pretty ok with each object having polices and views, things like paused scenes don't receive events, scenes have `.sprites` and `.camera`.

We should also add a `walk()` utility function.

pathunstrom:
Number so we can separate discussion as appropriate:

1. I can make it a second mix in, maybe? Either way, it'll end up futzing with EventMixin since `__event__` is our entry point.
2. What do you mean query/filter options? Like `Scene.get`?
3. I definitely consider all scenes the engine's parent, but for the sake of the event mixin, it's just the subsystems and the top scene that counts. I'm not planning on changing how the engine stores them, just how it hands them off for publication.
4. Camera is stored along side sprites, we just gave that magic view in `.camera`.
5. What would `.walk` do exactly?

astronouth7303:
[Classes] It was mostly "I didn't think we were making systems and sprites containers", so if we assume anything can have children, we might have to define some semantics.

[Querying] Basically? Kind of an ORM-ish thing? Like `myscene.sprites.filter(tag='spam')`

[Engine] Yeah? I was just thinking that `Engine.children` should include all scenes, as well as an `Engine.scenestack`.

`walk()` would just be a generator that produces an object and all its decedents.

pathunstrom:
Classes: Fair point. Going to definitely separate this and make it a subclass of EventMixin so it can be applied to scenes and the engine separately.

Engine: I dunno. If we do `engine.children` including all of it, we have to do extra checks to exclude non-active ones from publishing. I'd rather just make it explicit "children, as far as event propagation is concerned, are these things" and on the engine make that "current_scene + subsystems". This might be an argument for renaming children to something else.

Walk could be implemented. I don't really see the usage right now, though?

pathunstrom:
This all seems to work, going to open a new PR to handle the walk functionality and possible new publishing stuff.

ppb/events.py
```diff
@@ -85,6 +89,30 @@ def __event__(self, bag, fire_event):
                     raise
 
 
+class TreeStructurePublisher(EventMixin, IterableABC):
+    """
+    A mixin to publish events to an iterable's children.
+    """
+
+    def __iter__(self):
```
astronouth7303: I'm not super stoked about reserving `__iter__` to define children. I would rather see a `__children__` or `__tree_children__`.

I know we don't use it for anything right now, but it's such a huge part of Python it seems like a bad idea to reserve it so no game object can use it for something more semantic.

ppb/events.py
```diff
@@ -85,6 +89,30 @@ def __event__(self, bag, fire_event):
...
+        return (x for x in ())
```
astronouth7303: This can also be `yield from ()`.

ppb/events.py
```diff
@@ -85,6 +89,30 @@ def __event__(self, bag, fire_event):
...
+
+    def __event__(self, bag, fire_event):
+        super().__event__(bag, fire_event)
+        for child in self.publish_targets:
+            child.__event__(bag, fire_event)
+
+    @property
+    def publish_targets(self):
```
astronouth7303: I'm kinda wondering if this should be a dunder method? I'm kinda feeling like our core mechanisms should be dunders, to avoid reserving names.

Also, making this a property is kinda meaningless. I can't imagine a situation where it's set once and never updated? A generator property in particular is... ooph.

ppb/events.py
```diff
@@ -25,6 +26,7 @@
     'SceneStarted',
     'SceneStopped',
     'StopScene',
+    'TreeStructurePublisher',
```
astronouth7303: Minor quibble: `EventMixin` and `TreeStructurePublisher` should probably be at the top.

ppb/engine.py
```diff
@@ -50,6 +50,12 @@ def __init__(self, first_scene: Type, *,
         self.systems = []
         self.exit_stack = ExitStack()
 
+    def __iter__(self):
+        for s in self.systems:
+            yield s
```
astronouth7303: `yield from`

ppb/engine.py
```diff
@@ -50,6 +50,12 @@ def __init__(self, first_scene: Type, *,
...
+        for s in self.scenes:
+            yield s
```
astronouth7303: `yield from`

ppb/engine.py
```diff
@@ -114,21 +120,30 @@ def signal(self, event):
...
+    def publish_targets(self):
+        """
+        Yields children who should receive event publication.
+
+        Expected results are all subsystems and the current scene.
+
+        Overrides the TreeStructurePublisher mixin publish_targets.
+        """
+        for s in self.systems:
+            yield s
```
astronouth7303: `yield from`

pathunstrom:
We decided against this pattern for now. May come back to it some time in the future.

PR #231, opened 2019-04-08
The scripts I use to manage uploading to PyPI.

pathunstrom:


requirements-upload.txt
```diff
@@ -0,0 +1,9 @@
+certifi==2018.8.24
```
astronouth7303: Could you reduce this to what's actually required, not just a `pip freeze`?

pathunstrom: Probably doable. Looks to me like it's literally just twine and up to date pip.

pathunstrom:
I think most of the clunky is that I'm intentionally adding manual checks to make sure it doesn't go terribly sideways.

PR #232, opened 2019-04-08
requirements: Specify which version of ppb-vector is compatible

nbraud:


nbraud:
So, should this go in?

astronouth7303:
No, `master` should actually use the ppb-vector RC, as soon as it becomes available.

nbraud:
@astronouth7303 The current version of `master` isn't compatible with the current (development version of) vector.  #204 can change the version dependency as appropriate.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 pygame
-ppb-vector
+ppb-vector == 0.3
```
pathunstrom:
```suggestion
ppb-vector < 1.0.0
```

This matches what I did in the 0.5.1 release.

astronouth7303:
Closing in favor immutable vector merging

PR #235, opened 2019-04-21
Adds new example and animated sprite.

pathunstrom:
The blob resources were created by Piper Thunstrom and released
cc-0 with the source of ppb.


examples/animated_sprite.py
```diff
@@ -0,0 +1,26 @@
...
+    target = ppb.Vector(0, 0)
+    speed = 1
+
+    def on_mouse_motion(self, event: events.MouseMotion, signal):
+        self.target = event.position
+
+    def on_update(self, event: events.Update, signal):
+        intent_vector = self.target - self.position
+        self.position += intent_vector.scale(self.speed * event.time_delta)
+        self.rotation = math.degrees(math.atan2(intent_vector.y, intent_vector.x)) + 90
```
astronouth7303: Why not `.facing`?

astronouth7303: Because it's not settable. Ok.

PR #237, opened 2019-04-23
Flip Y Axis

astronouth7303:
Per some discussions on discord, flip the Y axis so that PPB's concept of space & coordinates match those of traditional geometry & algebra.

I think I'm going to write a full blog post explaining this.


astronouth7303:
Oh, this depends on #204.

ppb/camera.py
```diff
@@ -96,10 +96,18 @@ def in_frame(self, sprite: BaseSprite) -> bool:
...
+        """
         offset = (point - self.viewport_offset) * (1/self.pixel_ratio)
-        return self.position + offset
+        loc = self.position + offset
+        return loc.update(y=-loc.y)
 
-    def translate_to_viewport(self, point:Vector) -> Vector:
+    def translate_to_viewport(self, point: Vector) -> Vector:
+        """
+        Converts a vector from in-game to PyGame coordinate space
```
pathunstrom: This is more about screen space, not pygame specific (We'll have  a similar pixel array context no matter which rendering library we use.)

astronouth7303: Umm... You don't have too? Taking to discord.

PR #239, opened 2019-04-25
[WIP] New Camera API

pathunstrom:
We don't like the old camera, it's fiddly and hard for users to reason about.
This new API focuses on the actual dimensions of the camera as the frame.
The viewport is defined in percentages of the physical system. This way users
can think in game space and let the engine handle translating back and forth.

Will be reasoning through this and adding tests to hit the major requirements.

### DIFF
--- a/tests/test_camera.py
+++ b/tests/test_camera.py
@@ -16,15 +16,17 @@ def test_camera_sets_height():
     # NotSet will be a flag so we have a sentinel value
     assert cam.height is None
 
-    scene_started_event = SceneStarted(None)
-    scene_started_event.aspect_ratio = 2  # The width should be twice the height.
-    # Not true in normal contexts, but easier to test.
-    cam.__event__(scene_started_event, lambda x: None)
+    # The renderer will call this and pass in the physical aspect ratio.
+    # The camera can use that ratio and it's viewport definition to determine
+    # its aspect ratio.
+    cam.set_aspect_ratio(2)
     assert cam.height == 5
 
 
 def test_camera_change_frame_dimensions():
     """Test setting the dimensions at various points in the Camera lifetime."""
+    # TODO: This looks more like a test suite to me.
+    # Will probably break them into individual functions later.
 
     # Setting the height before the camera knows the aspect ratio.
     cam = Camera()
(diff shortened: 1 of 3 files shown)


tests/test_camera.py
```diff
@@ -1,8 +1,93 @@
...
+
+def test_camera_sets_height():
+    """Guarantee that the height is initially set dynamically."""
+
+    cam = Camera()
+
+    # Going to keep the default frame of 10 game units wide.
+    assert cam.width == 10
+    # NotSet will be a flag so we have a sentinel value
+    assert cam.height is None
```
astronouth7303: None is the sentinel value for this?

tests/test_camera.py
```diff
@@ -1,8 +1,93 @@
...
+
+    scene_started_event = SceneStarted(None)
+    scene_started_event.aspect_ratio = 2  # The width should be twice the height.
```
astronouth7303: This potential changes a lot? Setting it as part of the rendering process makes more sense?

So, my thought is, rendering can't be completely decoupled from cameras, and broadcasting the geometry doesn't seem super useful? Directly setting the geometry of each camera each frame still seems like the simplest, less weird way to do things.

tests/test_camera.py
pathunstrom: I had it as a flag, but realized None is always going to be an invalid value for this use case.

That said. . . the rest of the engine expects those sentinel flags, so should probably do the same here.

tests/test_camera.py
pathunstrom: I was thinking having the renderer extend scene started.

But per my comment on discord: I can probably do this magic on heartbeat events to reduce the time delta between any camera getting instantiated or screen (or eventually viewport) size changing and the Camera being able to manage itself. And let the renderer just call a method or set a property on the camera directly.

tests/test_camera.py
astronouth7303: None is python standard for "there's no value here". To be honest, replacing DoNotRender with None is perfectly reasonable?

pathunstrom: If we go that route, we should replace DoNotRender with something like a Square flag. In that case, we have two separate sentinel values: one will be the canonical way to remove an object from rendering, and the other needs to be the canonical way to do a drawn primitive.

Probably a conversation to take to a different issue, since it's a problem to be solved when we handle the different rendering contexts (Specifically, Text) thought I think the Animation system gave us a pretty good hint in that direction.

pathunstrom:
This is old and a lot has happened since then. Closing this PR.

PR #240, opened 2019-04-25
CI: Hardcode pip requirements into the Dockerfiles

nbraud:
Pro: Works around an issue with the way Cirrus caches containers: cirruslabs/cirrus-ci-docs#273

Cons:
- Requires regenerating/recommiting Dockerfiles whenever requirements change.
- This will break if the character `'` (rather than `"`) is used in a requirements file.

Marked as part of project [vector-1.0](https://github.com/orgs/ppb/projects/6) as it is blocking for #204.

### DIFF
--- a/.ci/dockerfiles/docs
+++ b/.ci/dockerfiles/docs
@@ -4,7 +4,6 @@ RUN apt update || true && \
     apt install -qq -y make && \
     rm -rf /var/cache/apt/*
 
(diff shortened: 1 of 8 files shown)


nbraud:
OK, Windows builds [fail](https://cirrus-ci.com/task/4616888431673344) due to some issue with how quoting works under Powershell...

PR #241, opened 2019-04-27
GameObjectCollection.add: Behave correctly with inheritance

nbraud:
- [x] Make scene tests also use a subclass of `TestPlayer`.
  This shows that `scene.get(kind=TestPlayer)` misses (instances of) subclasses.
- [x] Fix the behaviour in `GameObjectCollection`.
  Done by adding to all kinds in the MRO.  Possibly not the most efficient solution.
- [x] Refactor `tests/scenes/get_methods`.

### DIFF
--- a/ppb/scenes.py
+++ b/ppb/scenes.py
@@ -48,7 +48,9 @@ class GameObjectCollection(Collection):
         if isinstance(tags, (str, bytes)):
             raise TypeError("You passed a string instead of an iterable, this probably isn't what you intended.\n\nTry making it a tuple.")
         self.all.add(game_object)
-        self.kinds[type(game_object)].add(game_object)
+
+        for kind in type(game_object).mro():
+            self.kinds[kind].add(game_object)
         for tag in tags:
             self.tags[tag].add(game_object)
 
--- a/ppb/scenes.py
+++ b/ppb/scenes.py
@@ -93,7 +93,8 @@ class GameObjectCollection(Collection):
             container.remove(myObject)
         """
         self.all.remove(game_object)
-        self.kinds[type(game_object)].remove(game_object)
+        for kind in type(game_object).mro():
+            self.kinds[kind].remove(game_object)
         for s in self.tags.values():
             s.discard(game_object)
 
--- a/ppb/scenes.py
+++ b/ppb/scenes.py
@@ -124,6 +124,14 @@ class BaseScene(Scene, EventMixin):
         return (x for x in self.game_objects)
 
     @property
+    def kinds(self):
+        return self.game_objects.kinds
+
+    @property
+    def tags(self):
+        return self.game_objects.tags
+
+    @property
     def main_camera(self) -> Camera:
         return next(self.game_objects.get(tag="main_camera"))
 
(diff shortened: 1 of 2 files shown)


nbraud:
I had missed a necessary change in `GameObjectCollection.remove()` (f64dc2b).
Now fixed, but the testsuite for `GameObjectCollection` is clearly not sufficient.

ppb/scenes.py
```diff
@@ -48,7 +48,9 @@ def add(self, game_object: Hashable, tags: Iterable[Hashable]=()) -> None:
         if isinstance(tags, (str, bytes)):
             raise TypeError("You passed a string instead of an iterable, this probably isn't what you intended.\n\nTry making it a tuple.")
         self.all.add(game_object)
-        self.kinds[type(game_object)].add(game_object)
+
+        for kind in type(game_object).mro():
```
pathunstrom: This is a tricky thing: mro here is going to include `BaseSprite`, `Sprite` (the ABC), and `object`. Are you intending for that to happen or would you rather explicitly exclude those?

tests/test_scenes.py
```diff
@@ -26,9 +30,15 @@ def containers():
...
 
-@fixture()
-def player():
-    return TestPlayer()
+def players():
+    yield TestPlayer()
+    yield TestSubclassPlayer()
+
+
+def players_and_containers():
```
pathunstrom: You can do this with `pytest.mark.parametrize`. [Notes at the end of the first section of the docs here](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions).

> To get all combinations of multiple parametrized arguments you can stack parametrize decorators:
> ```python
> import pytest
> 
> 
> @pytest.mark.parametrize("x", [0, 1])
> @pytest.mark.parametrize("y", [2, 3])
> def test_foo(x, y):
>     pass
> ```
> 
> This will run the test with the arguments set to x=0/y=2, x=1/y=2, x=0/y=3, and x=1/y=3 exhausting parameters in the order of the decorators.

nbraud: This is explicitely not `parametrize` because it needs to be a *different* container for each test (as opposed to reusing the container across tests)

ppb/scenes.py
nbraud: I couldn't think of any reasons to exclude them (except maybe `object`?), and there are no documented exceptions about what are valid `kinds`.

tests/test_scenes.py
pathunstrom: Going to think on that, then. Ignore for now.

nbraud:
> I'd rather avoid using the mro in the tests themselves

What do you mean?  The tests do not currently use `mro()`.

pathunstrom:
They don't, but that's also the "easy" way to write the `remove` test. Was mostly saying not to go that route.

ppb/scenes.py
pathunstrom: The question is: what's the difference between `self.all` and `self.kinds[object]` under this new functionality?

nbraud:
Done.  I'm not particularly ameObjectCollection` testsuite, but fixing it is way out of scope here.

nbraud:
OK for merging this as-is, then?

PR #244, opened 2019-04-29
WiP: Main loop instrumentation

nicoonoclaste:
- [x] Optionally collect per-frame statistics in the main loop.
- [x] Pull in a somewhat-representative example (`mutant-games/hugs`)
- [x] Add automatic-play ability to the example, produce main loop statistics files.
  For simplicity, I did it by defining an `AISprite` variant of `PlayerSprite` that implements custom behaviour, rather than simulating input events.
- [x] Improve automatic play by computing trajectory intercepts (see commit msg in 93ccbc0933719dd0faa7e30b3a6d32c9fcd2a4d1)
  This is entirely unnecessary, but I somewhat nerdsnipped myself.
- [x] Process the per-frame statistics into more-useful data (convert to actual runtimes, ...)
- [x] Visualise the per-frame statistics
- [x] Extend statistics to include GC activity


nbraud:
Does the suggested instrumentation design makes sense?

I pulled in pandas, which is a pretty heavy dependency (pulls in numpy and a bunch of other things), which I believe to be reasonable because:
- this is entirely optional (only `import`ed when producing statistics) and only useful for developers;
- the data analysis script would likely depend on it (or something similar) anyhow, since I'm so not NIHing a statistics & tabular computations package  :3

nbraud:
Ohno, the nanosecond counters only exist in Python 3.7 and later  :(
I guess I should revert to the float counters?

nbraud:
OK, now that I have visualisation & stats output, I made the most basic control loop for the engine, and there is already a 12-fold improvement in jitter (0.9ms RMS -> 0.075ms RMS) and a large CPU usage reduction (we now spend the vast majority of our time sleeping).

I would suggest the following changes to the PR before merging:
- [ ] Make `engine.main_loop(collect_statistics=True)` return a sequence of statistics data objects, remove all (optional) dependencies on pandas from the engine, and instead convert to a pandas dataframe when saving the statistics.
- [ ] Move `main_loop_statistics.py` from `examples/` to an optional feature, so it can be installed (along with dependencies) by installing `ppb[stats]` or somesuch.
- [ ] Start documenting in `docs/discussion` how the main loop's timing is controlled.

I'd like feedback from @ppb/maintainers before doing all that, though.

Next steps after this PR could be adding GC scheduling (right now it's just being run every 100th iteration), so we can avoid running the GC when short on time, or when we predict no garbage is being produced.

### Without control loop (runs as fast as possible) 
interframe_jitter
```
signal for 2.07902856865076us, std. dev. 1.2948423539554727us
events for 227.00274972899868us, std. dev. 262.62686745205524s
gc for 87.50765258101441us, std. dev. 870.4983390314795us
sleep for 2.5642694571424347us, std. dev. 1.3434179359617922us

Output frame every 0.319141841796805ms (3133.402985863232 fps), std. dev. 0.9098870778953516ms
```

### With control loop (aimed at ~120fps)
interframe_no_jitter
```
signal for 11.737676056137905us, std. dev. 5.277532576202894us.
events for 1234.130141274278us, std. dev. 1063.7694681388045us.
gc for 0.42927634756347627us, std. dev. 0.4182789372143108us.
sleep for 7161.184066404758us, std. dev. 1054.3064573523095us.

Output frame every 8.405620852343532ms (118.9680117110201 fps), std. dev. 0.07588957643094331ms
```

PR #247, opened 2019-05-04
Revert "Merge #225"

astronouth7303:
This rolls back the use of CI dockerfiles, so that cirrus brokenness no longer blocks other PRs.

This is not a rejection of #240.


nbraud:
Thanks for doing that  <3

I need to revisit #240, but TBH I've been too frustrated with the quoting mess...  :(

PR #255, opened 2019-05-06
Allow external loop

moshez:


ppb/engine.py
```diff
@@ -95,12 +95,18 @@ def start(self):
...
             time.sleep(0)
-            now = time.monotonic()
-            self.signal(events.Idle(now - self._last_idle_time))
-            self._last_idle_time = now
-            while self.events:
-                self.publish()
-            self.manage_scene()
+            self.run_one_iteration()
+
+    def run_one_iteration(self):
```
nbraud: `run_one_iteration` isn't a very clear name; perhaps `run_one_frame`?

pathunstrom: That's possibly deceptive. One iteration of the event loop does not correspond to one animation frame or simulation frame anymore.

moshez: Maybe `run_one_loop_iteration`? The name is important here, because this is the public API that will be exposed to non-native event loops.

moshez:
```
pathunstrom Today at 7:12 AM
I'll make a command decision, with the note that we're still unstable and can fix it later, let's call it loop_once and have it working.
```

moshez: done

ppb/__init__.py
```diff
@@ -23,13 +34,10 @@ def run(setup: Callable[[BaseScene], None]=None, *, log_level=logging.WARNING,
...
-    kwargs = {
-        "resolution": (800, 600),
-        "scene_kwargs": {
-            "set_up": setup,
-        },
-        "window_title": title,
-    }
-
-    with GameEngine(starting_scene, **kwargs) as eng:
+    with GameEngine(starting_scene, **_make_kwargs(setup, title)) as eng:
```
astronouth7303: Why not just call `make_engine()`?

moshez: Mostly because it didn't seem like it would make the code any shorter or clearer...

moshez:
I am not sure where to make tests. I tested `loop_once` manually by running the `examples/targets_with_twisted.py` and confirming I can shoot the targets.

moshez:
added a test

PR #262, opened 2019-05-06
"Resizing errors"

kangaroo42:


ppb/systems/__init__.py
```diff
@@ -82,6 +82,8 @@ def prepare_resource(self, game_object):
             self.register_renderable(game_object)
 
         source_image = self.resources[image_name]
+        if game_object.size <= 0:
+            return None
```
astronouth7303: I would prefer this up 4 lines, but this arguably just fine.

PR #264, opened 2019-05-08
New documentation organization

pathunstrom:
Mentioned in the slack channel, now I'm implementing it. We can improve the language and various other features going forward. (Also, titles.)


astronouth7303:
Did we want to keep discussion in the blog? Or stick with docs?

pathunstrom:
My thought is that some of it (like underlying assumptions about the engine) are "timeless" and probably not really great for the blog, since blogs are inherently time sensitive. Things that aren't like that (like when we're discussing between alternatives and have chosen one) probably belong on the blog.

docs/reference/index.rst
```diff
@@ -0,0 +1,17 @@
+API Reference
+================
+
+For as simple as the tutorials make ``ppb`` look there's a lot of power under
+the hood. This section will cover the raw what of the ppb API. To find out why
```
nbraud: > the raw what of the ppb API.

That's odd/confusing phrasing, IMO.
Not a blocker though.

docs/tutorials/index.rst
```diff
@@ -0,0 +1,7 @@
+Tutorials
+============
+
+Tutorials live here, except for the basic Quick Start tutorial.
```
nbraud: Why the exception, re: quick start? 
It could live here and be linked directly from the front page.

Or is the quick start intrinsically not a tutorial?  In that case we shouldn't mention it.

pathunstrom: Mostly: Much smaller scope, also, I want it in the table of contents on the main page, which unless someone else knows how to do magic like that, I think means it needs to live a level up.

docs/reference/index.rst
pathunstrom: All of these literally exist as stubs, so yeah, better phrasing all around eventually.

docs/tutorials/index.rst
astronouth7303: The sphinx toctree is actually one thing, spread out over multiple files.

You can see it in the current docs by looking at the current index, source and rendered. Both inline headings and subpages are included in the TOC.

PR #266, opened 2019-05-09
Getting Started Guide

pathunstrom:
A new getting started guide. Also adds a new example: getting_started, which is a game similar to targets, but matches the code of the tutorial exactly.


docs/getting-started.rst
```diff
@@ -262,5 +262,28 @@ add the new lines.
 
 Now, when you press the space bar, projectiles appear. They only appear once
 each time we press the space bar (unless your keyboard supports key repeats,
-then you'll see it shooting every so often). Now we need something to hit with
-our projectiles!
\ No newline at end of file
+then you'll see it shooting every so often). Next we need something to hit with
```
astronouth7303: Is keyboard repeating something PyGame does not filter for us?

pathunstrom: If it's happening at the hardware level, I don't think it can tell? Most keyboards don't have it enabled by default, but I've seen it appear in weird cases.

astronouth7303: It's easy to filter out. Don't allow another key press event until a matching key up event.

Do we have any idea what configurations this happens in? Define "weird"?

pathunstrom: Specifically, some keyboards don't do "hold" they throw new keypress events on a repeat. Some accessibility tools do it as well. I would rather we _don't_ filter it so we don't end up with our stuff shutting out accessibility features.

astronouth7303: That particular discussion is a tangent, although one we should maybe discuss?

I would like to just drop the parenthetical, on the grounds that it's better to be accessible than 100% correct. But I also don't know the first thing about writing tutorials, so feel free to ignore me.

pathunstrom: Mostly want it there for the folks who might have a weird setup for any reason, but we can drop it and when someone encounters it, I know how to put it back.

docs/getting-started.rst
```diff
@@ -0,0 +1,308 @@
...
+
+``main.py``::
+
+   class Projectile(ppb.BaseSprite):
+      size = 0.25
+      direction = ppb.Vector(0, 1)
+      speed = 6
+
+      def on_update(self, update_event, signal):
+          if self.direction.x and self.direction.y:
```
astronouth7303: Vector has a bool

docs/getting-started.rst
```diff
@@ -0,0 +1,308 @@
...
+
+If we wanted to, we could pull out this ``on_update`` function into a mixin that
+we could use with either of these classes, but I'm going to leave that as an
+exercise to the reader. Just like the player, we can put a square image in the
+same folder with the name ``projectile.png`` and it'll get rendered, or we can
+let the engine make a colored square for us.
+
+Let's go back to our player class. We're going to add a new button to the class
+attributes, then update the ``on_key_pressed`` method. Just like before, I've
+removed some code from the sample, you don't need to delete anything her, just
```
pathunstrom: `here` is misspelled

docs/getting-started.rst
```diff
@@ -0,0 +1,308 @@
...
+          self.position += direction * self.speed * update_event.time_delta
+
+If we wanted to, we could pull out this ``on_update`` function into a mixin that
+we could use with either of these classes, but I'm going to leave that as an
+exercise to the reader. Just like the player, we can put a square image in the
+same folder with the name ``projectile.png`` and it'll get rendered, or we can
+let the engine make a colored square for us.
+
+Let's go back to our player class. We're going to add a new button to the class
+attributes, then update the ``on_key_pressed`` method. Just like before, I've
```
pathunstrom: Should clarify that we're not adding a new method, just adding code to the existing method.

docs/getting-started.rst
```diff
@@ -0,0 +1,308 @@
...
+       direction = ppb.Vector(0, 0)
+       speed = 4
+
+       def on_update(self, update_event, signal):
+           self.position += self.direction * self.speed * update_event.time_delta
+
+This new ``Player`` moves a certain distance based on time, and a direction
+vector and its own speed. Right now, our direction is not anything (it's the
+zero-vector), but we'll change that in a moment. For now, go ahead and run the
+program a few times, changing the parameters to Vector and the speed and see
```
pathunstrom: Clarify that this should be the direction attribute that needs to be modified.

pathunstrom:
Please don't merge, got a few more changes to make.

docs/getting-started.rst
astronouth7303: "update" seems find for that?

astronouth7303:
lgtm, bors r+

astronouth7303:
lgtm

bors r+

astronouth7303:
bors retry

astronouth7303:
bors retry

astronouth7303:
bors retry

PR #271, opened 2019-05-14
WIP Magic Fields

### LINKED ISSUE #234: Magical Field System
Need a way to do fields such that:

* Allow conversions/castings when set
* Allow virtual/generated fields
* Allows values to be set in the class definition of a subclass

astronouth7303:
Implement a field system, to try to remove warts around things like:

```
class MySprite(BaseSprite):
    position = 0, 1
```

```
class MySprite(BaseSprite):
    facing = 1, 1
```

```
class MySprite(BaseSprite):
    left = -4
```

Closes #234

### DIFF
--- a/ppb/fields.py
+++ b/ppb/fields.py
@@ -59,7 +59,6 @@ def _build_fields_dict(cls, fieldbag):
         rv = dict(_iter_nonspecial_props(fieldbag))
 
     # Mask off descriptors, so field mechanisms don't prevent it from working
-    print(cls, vars(cls), list(_iter_nonspecial_props(cls)))
     for name, value in _iter_nonspecial_props(cls):
         if _is_descriptor(value):
             rv[name] = None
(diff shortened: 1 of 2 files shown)


ppb/fields.py
```diff
@@ -0,0 +1,91 @@
...
+    Converts an annotations dict into a fields dict
+    """
+
+
+class FieldMixin:
+    """
+    Mixin that implements all the field magic
+    """
+    def __init_subclass__(cls, **_):
+        # If new fields are defined, make a __fields__
```
pathunstrom: Should we worry about the lack of super call?

astronouth7303: I probably just forgot about it.

astronouth7303:
I think this is ready for the first round of comments.

In particular, I'm worried test coverage isn't good enough.

astronouth7303:
Oh, look, `ppb_mutant.index` does change things on the class level and expects that to work.

_sigh_ I'll have to implement that. Will probably refactor everything into a metaclass.

astronouth7303:
Since Piper and I both agree this might be a step _too_ magical, I'm going to close this PR.

PR #279, opened 2019-05-24
FauxFloat: Redo operations to allow for double-fauxs to work

astronouth7303:
Found that things like `min(s.left for s in scene.sprites)` didn't work.

Fixed that.

### DIFF
--- a/tests/test_fauxfloat.py
+++ b/tests/test_fauxfloat.py
@@ -45,11 +45,30 @@ def test_unary_ops(operation, num):
     num=st.floats(allow_nan=False, allow_infinity=False),
     other=st.floats(allow_nan=False, allow_infinity=False),
 )
-def test_binary_ops(operation, num, other):
+def test_binary_ops_float(operation, num, other):
     t = get_thingy(num)
 
     assert operation(t, other) == operation(num, other)
     assert operation(other, t) == operation(other, num)
+
+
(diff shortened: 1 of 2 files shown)


tests/test_fauxfloat.py
```diff
@@ -45,13 +45,32 @@ def test_unary_ops(operation, num):
...
+    [
+        operator.lt, operator.le, operator.eq, operator.ne, operator.ge,
+        operator.gt, operator.add, operator.mul, operator.sub,
+    ],
+)
+@given(
+    num=st.floats(allow_nan=False, allow_infinity=False),
+    other=st.floats(allow_nan=False, allow_infinity=False),
+)
+def test_binary_ops_thingy(operation, num, other):
```
nbraud: I sent a better proposal as #281.

nbraud: PS: This is missing a `test_binary_ops_thingy_nonzero`, but please don't actually do this  >_>'

nbraud: PS: This is missing a `test_binary_ops_thingy_nonzero`, but please don't actually do this  >_>'

nbraud:
@astronouth7303 I guess this means you approve of the testsuite changes?

bors r=astronouth7303,nbraud

PR #280, opened 2019-05-24
[WIP] Update for ppb_vector 1.0a2

astronouth7303:
Update for the vector breaking changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 pygame
-ppb-vector >= 1.0a1, < 2
+ppb-vector >= 1.0a2, < 2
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
 pygame
-ppb-vector >= 1.0a1, < 2
+ppb-vector >= 1.0a2, < 2
```
nbraud: This can be simplified by specifying a version [compatible with] 1.0a2:
```suggestion
ppb-vector ~= 1.0a2
```

[compatible with]: https://www.python.org/dev/peps/pep-0440/#compatible-release

PR #284, opened 2019-05-25
Remove abc.py

astronouth7303:
It was providing no value and hasn't been updated in several versions.


nbraud:
@astronouth7303 I rebased your PR to deal with a merge conflict in `events.py`

ppb/events.py
```diff
@@ -87,6 +83,12 @@ def __event__(self, bag, fire_event):
 
 # Remember to define scene at the end so the pargs version of __init__() still works
 
+from ppb.scenes import BaseScene
```
nbraud: Why are you moving the imports from the top of the file?

ppb/events.py
```diff
@@ -127,23 +129,23 @@ class StartScene:
         * `signal(new_scene=StartScene(MyScene(player=player))`
         * `signal(new_scene=StartScene, kwargs={"player": player}`
     """
-    new_scene: Union[Scene, Type[Scene]]
+    new_scene: Union[BaseScene, Type[BaseScene]]
```
nbraud: This is a confusing type signature.  Why is it either a subclass of `BaseScene` or an instance thereof?

ppb/scenes.py
```diff
@@ -99,15 +96,16 @@ def remove(self, game_object: Hashable) -> None:
...
     background_color: Sequence[int] = (0, 0, 100)
     container_class: Type = GameObjectCollection
 
-    def __init__(self, engine, *,                 
-                 set_up: Callable=None, pixel_ratio: Number=64,
+    def __init__(self, engine, *,
+                 set_up: Callable = None, pixel_ratio: Number = 64,
                  **kwargs):
-        super().__init__(engine)
+        super().__init__()
```
nbraud: Remove the call to `super().__init__`, `EventMixin` has no `__init__` method.

ppb/events.py
pathunstrom: Removing the ABC means we need another type for the type hints, and BaseScene depends on `ppb.events` which now depends on `ppb.scenes`. So a known trick for allowing circular imports.

Long term, I think we need to move `EventMixin` or the event definitions elsewhere so we're not doing this if we can avoid it.

ppb/events.py
astronouth7303: That's exactly what's going on.

The Engine either accepts a pre-constructed Scene instance, or it accepts a scene class and will construct it for you.

ppb/scenes.py
astronouth7303: Technically not true, because of C3 linearization.

However, I would be pretty ok with removing the `engine` mandatory positional argument and therefore making the `__init__` signature up and down the tree more predictable.

ppb/events.py
nbraud: :+1: on breaking the circular dependency.

ppb/events.py
nbraud: OK, why is it `Type[BaseScene]` rather than `Callable[..., BaseScene]` then?  (i.e. shouldn't we allow any sort of callable there, not only class constructors?)

ppb/scenes.py
pathunstrom: I think the engine mandatory argument is a thing that's been on my list to kill for a couple of versions now, do so if you'd like.

ppb/events.py
astronouth7303: I think it actually checks for a `type` to determine if it should call it.

ppb/scenes.py
astronouth7303: Ok, see https://github.com/astronouth7303/pursuedpybear/pull/2

ppb/events.py
nbraud: Thanks for all the explanations  <3
Changing that seems out-of-scope here, but should I make an issue for it?

astronouth7303: I'm not sure?

Basically, you need to determine the right shape to check for, and the obvious choices are:

* Checking for instances of `type` (disallowing non-class factories)
* Checking for callables (disallowing callable scenes)

Neither of these are super-great trade-offs? But also, neither of them are encouraged patterns.

nbraud: Fair.

PR #286, opened 2019-05-28
Run linters during CI

nicoonoclaste:
This adds a `lint` task in CI, which uses the same linters & configuration as `ppb-vector`.
It is a follow-up on #285, which fixes the warts found by those linters.

- `flake8`;
- [`flake8-bugbear`], which looks for likely bugs and design problems;
- `flake8-commas`, for enforcing trailing commas in multi-line dicts, lists, tuples, ...
- `flake8-import-order`, for enforcing the import conventions we already use;
- `pep8-naming`, for enforcing [PEP8]'s naming conventions.

In the few places where it was necessary, `noqa` annotations were used, scoped to a specific lint.

[PEP8]: https://www.python.org/dev/peps/pep-0008/
[`flake8-bugbear`]: https://github.com/PyCQA/flake8-bugbear


gideongrinberg:
@nbraud does it make sense to lint builds or to use [pre-commit hooks](https://pre-commit.com/)? These will prevent poor code from being committed. You can also set up pre-commit to run a formatter like black or autopep8 (although I know there are mixed opinions about formatters.) I'm not sure exactly what the differences are, but it might be worth looking at. It probably comes down to performance and convenience.

AstraLuma:
Only if people have installed the additional tools. We have a strong desire to be as friendly as possible for people to get started contributing.

The problems with this aren't "how to run the tool" but "strategy for bringing the code base into compliance".

gideongrinberg:
> Only if people have installed the additional tools. We have a strong desire to be as friendly as possible for people to get started contributing.

That makes a lot of sense Astra. The thing is (and this is just my opinion) linting *before* you push is important. It enforces good habits, makes the dev focus on readability, and makes code review easier. While it could increase the barrier to entry, most people are just running `pip install -r requirements.txt` anyway and pre-commit works automatically. It could make it harder for beginners to write code but, as I said, good quality should be enforced. What would be your opinion on formatting in the CI builds?

AstraLuma:
The ci check stays.

If the tool is that automatic, can you open an issue for us to discuss it? I want to make sure I understand how well it works in various scenarios.

gideongrinberg:
@AstraLuma see #531

PR #289, opened 2019-06-06
[nomerge] Nix __resource_path__

astronouth7303:
This just removes `__resource_path__` and makes all resources resolution relative to the Python module "virtual filesystem".

Note that this requires an `__init__.py` to exist next to any resource files.

### DIFF
--- a/ppb/vfs.py
+++ b/ppb/vfs.py
@@ -1,0 +1,60 @@
+"""
+Handles opening files from the Python "VFS".
+"""
+try:
+    import importlib.resources as impres
+except ImportError:
+    # Backport
+    import importlib_resources as impres
+
+
+def _splitpath(filepath):
+    if '/' in filepath:
+        slashed, filename = filepath.rsplit('/', 1)
+        modulename = slashed.replace('/', '.')
+    else:
+        modulename = '__main__'
+        filename = filepath
+
+    return modulename, filename
+
+
+def open(filepath, *, encoding=None, errors='strict'):
+    """
+    Opens the given file, whose name is resolved with the import machinery.
+
+    If you want a text file, pass an encoding argument.
+
+    Returns the open file and the base filename (suitable for filename-based type hinting).
+    """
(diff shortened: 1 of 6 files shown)


ppb/vfs.py
```diff
@@ -0,0 +1,97 @@
+"""
+Handles opening files from the Python "VFS".
+"""
+import logging
+from pathlib import Path
+import sys
+try:
+    import importlib.resources as impres
+except ImportError:
+    # Backport
```
nbraud: Please specify in a comment which Python versions require that.  (Presumably, only 3.6?)

ppb/vfs.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def _main_path():
+    main = sys.modules['__main__']
+    mainpath = getattr(main, '__file__')
+    if mainpath:
+        mainpath = Path(mainpath)
+        return mainpath.resolve().parent
+    else:
+        return Path.cwd()
```
nbraud: Document when `Path.cwd()` is expected to be used, and why it's a good choice.

ppb/vfs.py
nbraud: Please specify in a comment which Python versions require that.  (Presumably, only 3.6?)

astronouth7303: Yeah, it's just 3.6

ppb/vfs.py
astronouth7303: This is loosely based on the deleted `__resource_path__` implementation, where under a REPL `__main__` will refer to the console namespace and not have a `__file__`. (It used [`getfile`](https://github.com/python/cpython/blob/3.7/Lib/inspect.py) though)

astronouth7303:
Note that I never really intended this to be merged, it's meant to be a demonstrative spike of "How would we do VFS?". I would much rather see the general resource revamp we've been discussing that incorporated this.

astronouth7303:
Closing in favor of #306

PR #290, opened 2019-06-07
Limit Python versions

astronouth7303:
Don't allow Python 3.8.

This is preemptive because PyGame 1.9 is currently not planned to be compatible with 3.8, and while both Python 3.8 and PyGame 2 are a ways off, we want to prevent accidental incompatibilities the best we can.


pathunstrom:
LGTM

bors r+

PR #295, opened 2019-06-18
Adds engine extension keyword argument

### LINKED ISSUE #263: Extensions API
One of the sour points of ppb right now is subsystems _are hard_ to add to the engine. You have to rebuild the `systems` kwarg to get the basic engine + your new thing.

Let's add a new kwarg that gets appended to the systems kwarg  for this use case.

We should discuss here whether that should be called `extensions` or `plugins` or something else.

- [ ] Pick a name.
- [ ] Add keyword argument `???`
- [ ] Modify engine to use new kwarg.
- [ ] Test it!
- [ ] Make sure name is in CONTRIBUTORS.md

pathunstrom:
Resolves #263.

Went with `basic_systems` and `systems`. Used pycharm refactoring tools to change all references to the old `systems` and update to the new `basic_systems`, then modified a test to require our new `systems` keyword argument, then combined using `chain` that was already in the `engine.py` namespace.

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -27,7 +27,7 @@ class GameEngine(EventMixin, LoggingMixin):
 
     def __init__(self, first_scene: Type, *,
                  basic_systems=(Renderer, Updater, PygameEventPoller),
-                 scene_kwargs=None, **kwargs):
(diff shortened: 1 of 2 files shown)


astronouth7303:
bors retry

PR #303, opened 2019-07-07
requirements.txt: Only accept stable releases of ppb-vector

nbraud:


nbraud:
@astronouth7303 Looks like you forgot to bump this during the release.
This might be worth issuing a v0.6.1 release.

PR #306, opened 2019-07-11
Revamp Assets

### LINKED ISSUE #147: Resource Subsystem
Add system to load and cache resources, including zip safe.

astronouth7303:
Totally revamp how PPB handles assets.

Depends on #313 because life was just easier that way.

Part of #147


ppb/systems/__init__.py
```diff
@@ -99,8 +101,9 @@ def prepare_rectangle(self, resource, game_object, camera):
 
     def register(self, resource_path, name=None):
```
pathunstrom: I feel like we probably want to rename the parameter here? Since it's not resource_path anymore?

astronouth7303: Oops! This code actually got totally nixed. I forgot to mark this PR as WIP.

tests/test_sprites.py
```diff
@@ -326,18 +326,3 @@ def test_rotatable_base_sprite():
 
     test_sprite.rotate(1)
     assert test_sprite.rotation == 1
-
-
-def test_sprite_in_main():
```
pathunstrom: Is there a solid way to keep a test around this? The use case of "no actual __file__" is still a possibility, we just resolve it differently.

ppb/systems/__init__.py
```diff
@@ -78,14 +98,14 @@ def render_background(self, scene):
...
-        image_name = game_object.__image__()
-        if image_name is flags.DoNotRender:
+        image = game_object.__image__()
+        if image is flags.DoNotRender:
             return None
-        image_name = str(image_name)
-        if image_name not in self.resources:
-            self.register_renderable(game_object)
+        if isinstance(image, str):
+            logger.warn(f"Using string resources is deprecated, use ppb.Image instead. Got {image!r}")
```
pathunstrom: 👍

ppb/systems/__init__.py
```diff
@@ -22,6 +26,22 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 
 from ppb.systems.pg import EventPoller as PygameEventPoller  # To not break old imports.
+from ppb.assets import Asset
+import io
+
+
+class Image(Asset):
```
pathunstrom: Dunno on adding new things to `systems.__init__`. If we're making changes to the renderer and adding new things they should move to its own file.

ppb/systems/__init__.py
astronouth7303: Only thing about this is that it's extremely verbose (every frame).

ppb/systems/__init__.py
astronouth7303: Yeah, I was wondering if we wanted to move the renderer to its own module.

pathunstrom: It's one of the places we deal with pygame, so probably needs to go to `systems.pg`. (There's even argument for making pg.py into a subpackage of its own, but uh. . . not right now.)

ppb/systems/__init__.py
pathunstrom: Could throw in a tracking variable and only throw it once? (Or once per object?) Would `lru_cache` wrapped around a print statement make it run once per object passed?

astronouth7303:
This still needs tests.

ppb/systems/__init__.py
astronouth7303: Yeah, all of those things. It's just annoying, though?

astronouth7303:
@pathunstrom I think this is ready for review & merge.

PR #307, opened 2019-07-11
Enable passing arbitrary options to the engine through the standard launchers

astronouth7303:
Punishing users for wanting to use features isn't a great look.

This allows users to use the standard startup functions with arbitrary options (like `systems`).


pathunstrom:
bors retry

PR #309, opened 2019-07-12
Adds a CHANGELOG.txt

pathunstrom:
The goal of this is to have a place to put notes when we merge in new features, change existing features, and remove features. This includes the notes from every previous release to 0.2.0 (0.2.0 is the release that restarted the project.)

Going forward, the changelog can be modified separately from the features (to prevent too many merge conflicts with people modifying the changelog with their changes) and we should do that sometime after bors is finished cleaning up.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,0 +1,91 @@
+Version 0.7.0
+
+Changed:
+* Only accept stable versions of ppb-vector
+
+Version 0.6.0
+
+New:
+* Added docs (ppb.readthedocs.io)
+* Sprite rotation
+* Idle event
+* Animation feature
+* Window Title added to run
+* ppb.make_engine added
+* ppb.loop_once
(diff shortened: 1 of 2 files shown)


astronouth7303:
I'm not super happy about "try to remember to update the changelog occasionally" as a process?

Also, is the changelog on the github releases page insufficient?

pathunstrom:
Surfacing off github discussion:

The point of this is less that the changelogs attached to releases are insufficient, it's more I'm tired of going back and having to dig through and often missing features because it's not something I was considering.

The goal is that whoever merges a feature opens a new PR immediately (or soon after) including those notes in this file.

astronouth7303:
Ok, to be clear:

Yes, compiling changelogs from commit or PR history is very annoying. It would be really nice if we could track changelog entries as we developed.

I'm not a fan of "remember to make a PR" as a process.

pathunstrom:
Closing until we figure out a solution.

PR #314, opened 2019-07-13
docs/scenes: Fix indentation

astronouth7303:


astronouth7303:
... That is one of the more surreal moments I've had digging through git.

I remember fixing a problem where some parts in a file were using 3-space indents but a few were using 4-space and.... where did it go?

astronouth7303:
Ok, actually got the indentation fix in there.

Also, the `py:currentmodule` helps with `:py:*:` links, since it (1) gives a location for the module in general and (2) gives context in resolving not-fully-qualified links.

astronouth7303:
bors retry

PR #315, opened 2019-07-13
Add AssetLoaded event

astronouth7303:
Fire an event when an asset finishes loading and becomes available.

Depends on #306, #316


astronouth7303:
This is currently so plagued with import loops that I'm starting to wonder if we should just disallow them altogether.

astronouth7303:
Ok, 3.6 is happier now. I just had to rewrite, like, 2 dozen imports.

PR #316, opened 2019-07-13
Move things, break loops

astronouth7303:
Grab a u-haul! :truck: 

Generally moving things and reducing imports to improve the import loop situation

* Break up "the machinery implementing a thing" and "a pile of things that use the thing"
* Prefer module imports to individual imports, so any unavoidable loops don't wreck everything.

Depends on #306 because I didn't feel like dealing with the merge conflict.

No docs changes because this doesn't touch anything publicly documented.


astronouth7303:
Note: I know this code is all pygame-specific, but the way I figure, separating it out like this should make it easier to bundle it up into `ppb.systems.pygame` and autodetect things later.

astronouth7303:
Ok, I think this is ready to go as soon as #306 merges.

ppb/eventlib.py
```diff
@@ -0,0 +1,67 @@
+"""
+The event machinery
+"""
+
+import logging
+import re
+
+__all__ = (
+    'EventMixin',
+)
```
pathunstrom: We should export BadEventHandlerException, too.

ppb/systems/inputs.py
```diff
@@ -3,11 +3,11 @@
...
-from ppb import Vector
-from ppb import events
-from ppb import keycodes as keys
-from ppb.systems import System  # TODO: Be aware of circular imports
-from ppb.systems import default_resolution
+from ppb_vector import Vector
+import ppb.events as events
+import ppb.keycodes as keys
+from ppb.systems.base import System
+from ppb.systems.renderer import DEFAULT_RESOLUTION
```
pathunstrom: Not blocking this, but we should probably move shared defaults to their own file somewhere.

astronouth7303: Possibly? If we had more than one?

I grouped it like this because resolution felt like a renderer concern, and the renderer is responsible for window management currently?

astronouth7303:
@pathunstrom can you take a second pass on this?

pathunstrom:
bors retry

astronouth7303:
bors retry

astronouth7303:
bors retry

pathunstrom:
bors retry

Issue #276, opened 2019-05-20
Layering (Render Order)


astronouth7303:
At some point, we're going to need the ability to have a well-defined render order. I see a few ways to do this:

* Layers: Scenes declare layers by name and order, and sprites declare what layer they're part of
* Z axis: Sprites grow a z-index, and sprites are sorted based on their z-index
* Ordered Sprites: Sprites in a scene become ordered and are rendered in that order. Scenes grow methods to reorder sprites.

My personal opinions: Layers are semantic and meaningful, but require a bit more work in the renderer. Z axis is simple and easily nuanced, but lacks in intrinsic meaning. Ordered Sprites seems awkward to work with and would be more challenging to implement with #229.

pathunstrom:
I've thought about adding layers as an additional index in Scenes, which can read off of a layer attribute when an object gets added or set explicitly via keyword.

Then it's a matter of ordering the output of `__iter__` to follow the layers in one order or the other.

pathunstrom:
Which I guess is. . . both a and b combined.

jamesBrosnahan:
How do you feel about making a layer data class that holds a z-index and
returns objects holding a reference to it. Then having a container class
that gets read from during rendering?
https://stackoverflow.com/questions/11891755/find-all-references-to-an-object-in-python
this link recommends the usage of gc.get_referrers(*objs) to find objects
references [useful if the layers should be seamless]

pathunstrom:
@jamesBrosnahan How does that interact with the scene split? Will objects in non-active scenes end up visible if we grab _all_ references?

jamesBrosnahan:
How do you feel about having a layer class that uses
gc.get_referrers(*objs) to look up the objects that reference a layer
during rendering?

https://stackoverflow.com/a/33398553

jamesBrosnahan:
The layer gets read by the render loop and returns a list of references to
objects that are part of that layer. The objects then need to be processed
to determine if they should not be rendered, or the scene could own their
layer container and processing can be skipped if the scene is disabled.

astronouth7303:
That all seems needlessly complex and brittle?

Like, to implement z-index, you just need `sprites = sorted(scene.sprites, key=lambda s: s.z_index)`

Named layers just need something like `sprites = sorted(scene.sprites, key=lambda s: scene.layers.index(s.layer))` (except not exactly, because there's more edge cases).

We don't need explicit layer objects and tracking references and stuff? Just some primitive values?

[...]

PR #322, opened 2019-07-15
Sprite Layer API

pathunstrom:
Implements a basic layering API.

`Sprites` can have a `layer` attribute that should be an integer, but due to implementation details could be a float.

Does not change the renderer as part of this change, will do so in a followup PR.

This is one step on the way to #276

Proposal:

* `BaseScene.sprite_layers()` which returns all sprites based on their layering order.
* Layers are simple attributes on `Sprites`.
* You can name layers via `BaseScene.named_layers` which will default to incrementing each layer number by 1.
* You can have greater control over your layer definition by providing `BaseScene.defined_layers` as Dict[str, int].
* There is a `BaseSprite.default_layer` which a Sprite without a layer attribute defaults to. It can be a string or an int as preferred by the end user.

### DIFF
--- a/tests/test_layering.py
+++ b/tests/test_layering.py
@@ -172,7 +172,7 @@ def test_default_layer_named():
             "something_else": 2,
             "another": 3
         }
-        _default_layer = "our_default"
+        default_layer = "our_default"
 
     scene = LayeredScene()
 
@@ -186,9 +186,9 @@ def test_default_layer_set_layer():
 
     class LayeredScene(scenes.BaseScene):
         named_layers = ["background", "mobs", "particles"]
-        _default_layer = "particles"
+        default_layer = "particles"
 
     scene = LayeredScene()
 
     scene.default_layer = "mobs"
-    assert scene.default_layer == 1
+    assert scene.default_layer == "particles"
--- a/tests/test_layering.py
+++ b/tests/test_layering.py
@@ -40,11 +40,11 @@ def test_change_layer():
     for sprite in ones:
         scene.add(sprite)
 
-    assert list(scene.sprite_layers())[0] == test_sprite
+    assert list(scene.sprite_layers())[0] is test_sprite
 
     test_sprite.layer = 2
 
-    assert list(scene.sprite_layers())[-1] == test_sprite
+    assert list(scene.sprite_layers())[-1] is test_sprite
 
 
 def test_layering_without_layer_attribute():
@@ -56,4 +56,4 @@ def test_layering_without_layer_attribute():
     for x in range(1, 6):
         scene.add(LayeredSprite(x))
 
-    assert list(scene.sprite_layers())[0] == test_sprite
+    assert list(scene.sprite_layers())[0] is test_sprite
--- a/tests/test_layering.py
+++ b/tests/test_layering.py
@@ -8,10 +8,11 @@ class LayeredSprite:
         self.layer = layer
 
 
-class NoLayer: pass
+class NoLayer:
+    pass
 
 
-def test_layering_attribute():
+def test_layering_attribute_ints():
 
     class LayeredScene(scenes.BaseScene):
 
@@ -21,7 +22,7 @@ def test_layering_attribute():
                 self.add(LayeredSprite(x))
 
     scene = LayeredScene()
-    for lower_sprite, higher_sprite in zip(scene, list(scene)[1:]):
+    for lower_sprite, higher_sprite in zip(scene.sprites(), list(scene.sprites())[1:]):
         if isinstance(lower_sprite, camera.Camera) or isinstance(higher_sprite, camera.Camera):
             continue
         assert lower_sprite.layer < higher_sprite.layer
--- a/tests/test_layering.py
+++ b/tests/test_layering.py
@@ -20,7 +20,7 @@ def test_layering_attribute():
 
         def __init__(self):
             super().__init__()
-            for x in range(5):
+            for x in range(-3, 3):
                 self.add(LayeredSprite(x))
 
     scene = LayeredScene()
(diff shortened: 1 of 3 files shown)


tests/test_layering.py
```diff
@@ -0,0 +1,164 @@
...
+    test_sprite.layer = 2
+
+    assert list(scene)[-1] == test_sprite
+
+
+def test_registered_layers():
+
+    scene = scenes.BaseScene()
+
+    scene.define_layer("background", 0)
```
astronouth7303: This API is against our declarative direction.

pathunstrom:
> So far, we've avoided putting too much logic in BaseSprite and BaseScene and I'm not sure I want that to change?

That's definitely true of BaseSprite, I think BaseScene is just a matter of the only functionality we've needed is container and iteration and lookups.

> (I assume you want BaseScene.__iter__() to produce sprites in layer order)

Yes. That's primarily a backwards compatibility thing: We don't have to change renderer logic by making sure `__iter__` is in layer order.

I'm going to ponder on the rest of this. I've got ideas.

astronouth7303:
ok, so here's the concrete things against implementing ordering in `__iter__()`:

* It makes assigning to properties have a side-effect (and not just a side-effect of "update this cached pre-compute")
* It really doesn't fit with the discussions we've had about non-sprite children

tests/test_layering.py
```diff
@@ -21,13 +22,13 @@ def __init__(self):
                 self.add(LayeredSprite(x))
 
     scene = LayeredScene()
-    for lower_sprite, higher_sprite in zip(scene, list(scene)[1:]):
+    for lower_sprite, higher_sprite in zip(scene.sprite_layers(), list(scene.sprite_layers())[1:]):
```
astronouth7303: `itertools.islice()` might help here.

tests/test_layering.py
```diff
@@ -37,62 +38,78 @@ def test_change_layer():
...
-    assert list(scene)[-1] == test_sprite
+    assert list(scene.sprite_layers())[-1] == test_sprite
 
 
-def test_registered_layers():
+def test_defined_layers():
 
-    scene = scenes.BaseScene()
+    class LayeredScene(scenes.BaseScene):
+        defined_layers = {
```
astronouth7303: I hate this name

(I make this comment fully aware of the importance of names in discussions.)

tests/test_layering.py
```diff
@@ -135,17 +159,22 @@ def test_named_layers_with_undefined_layer():
 
     assert list(scene)[0] is test_sprite
 
-    scene.define_layer("undefined_layer", 6)
+    scene.defined_layer["undefined_layer"] = 6
```
astronouth7303: Is this name just a typo?

pathunstrom: Yup, should be `defined_layers`

tests/test_layering.py
pathunstrom: Um. . . `named_layers` maybe? It invites confusion with `layer_names` though.

tests/test_layering.py
pathunstrom: Instead of the list > slice thing? Yeah, I can make that change.

tests/test_layering.py
astronouth7303: I figured just `layers`?

tests/test_layering.py
```diff
@@ -0,0 +1,74 @@
...
+
+    class LayeredScene(scenes.BaseScene):
+
+        def __init__(self):
+            super().__init__()
+            for x in range(5):
+                self.add(LayeredSprite(x))
+
+    scene = LayeredScene()
+    for lower_sprite, higher_sprite in zip(scene.sprite_layers(), islice(scene.sprite_layers(), 1, None)):
```
astronouth7303: should `.sprite_layers()` include the camera at all?

pathunstrom: The primary point of sprite_layers is things that should be rendered, so. . . probably not? We can probably exclude both DoNotRender (and whatever follows it) and size 0 objects.

tests/test_layering.py
```diff
@@ -0,0 +1,74 @@
...
+
+    assert list(scene.sprite_layers())[-1] == test_sprite
+
+
+def test_layering_without_layer_attribute():
+
+    test_sprite = NoLayer()
+    scene = scenes.BaseScene()
+
+    scene.add(test_sprite)
```
astronouth7303: Umm... i know this is just tests, but wow, i didn't know scenes support non-sprite non-....anything

tests/test_layering.py
```diff
@@ -0,0 +1,74 @@
...
+    for x in range(1, 6):
+        scene.add(LayeredSprite(x))
+
+    assert list(scene.sprite_layers())[0] == test_sprite
```
astronouth7303: Should probably be `is`

tests/test_layering.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_set_default_layer():
+
+    class DefaultLayer(scenes.BaseScene):
+        default_layer = 3
```
astronouth7303: Is this worth allow customizing?

tests/test_layering.py
astronouth7303: Ok, should it be the scene's or the renderer's responsibility to manage render order and filter for renderability?

tests/test_layering.py
pathunstrom: This is only possible because it's not touching the renderer.

tests/test_layering.py
pathunstrom: From discord:

Layered iteration is a scene concern, renderability is a renderer concern.

Also, Camera should stop being a sprite. To be worked out in a different PR.

pathunstrom:
An implementation! Once this is merged, I can start work on both a layering mixin for Sprites and start the work on the Renderer rebuild.

tests/test_layering.py
```diff
@@ -0,0 +1,59 @@
...
+
+    test_sprite = LayeredSprite(0)
+    ones = tuple(LayeredSprite(1) for _ in range(3))
+
+    scene = scenes.BaseScene()
+    scene.add(test_sprite)
+    for sprite in ones:
+        scene.add(sprite)
+
+    assert list(filter(lambda s: not isinstance(s, camera.Camera), scene.sprite_layers()))[0] is test_sprite
```
astronouth7303: Should this just be a comprehension? And/or use `next()`?

astronouth7303:
One last comment: Did we want to add `layer: int = 0` to the base sprite class?

pathunstrom:
That's reasonable. Will add it.

pathunstrom:
bors retry

PR #323, opened 2019-07-16
internize assets

astronouth7303:
Multiple `Asset`s referring to the same file will be intern'd into the same instance.


astronouth7303:
I did just notice that the `AssetLoadingSystem` keeps futures forever, including the raw data, which kinda wrecks the caching model and memory usage. :/

astronouth7303:
bors retry

PR #325, opened 2019-07-16
broken links to tutorials and events

milasp:


astronouth7303:
This isn't broken on [RTD](https://ppb.readthedocs.io/en/master/getting-started.html)? And testing the change locally produces weird results.

Screenshot from 2019-07-16 13-46-52

pathunstrom:
Closing, combination of no action and the fact that the role in Sphinx is `:doc:` so this fix would actually break functioning documentation.

PR #326, opened 2019-07-16
Sound! 🔊

astronouth7303:
Add a sound system!

Part of #327


astronouth7303:
So, a quirk: pygame has a limited number of parallel sounds (channels) that can be played at once, defaulting to 8.

My initial thought was to just allocate more channels if we run out, but then I realized that 1. Most naive implementations would spam this easily, 2. It's not obvious when to scale down.

pathunstrom:
I figure we pick a number we think is reasonable and manageable and go with that. We can probably surface a parameter in the engine to modify it as well?

astronouth7303:
Yeah, a parameter is easy enough. The pygame mixer surfaces so many parameters, though, and tbh I don't think game developers should have to think about that?

pathunstrom:
I agree. We can pick the ones that are important and let advanced users subclass and modify as they like.

astronouth7303:
Reorganized things so sound is broken up across several PRs.

PR #329, opened 2019-07-20
Write an ABC for assets

astronouth7303:
Because of the confusion around assets, I figured I'd write an ABC about it.


pathunstrom:
Is this ready for review?

PR #334, opened 2019-07-28
imports 'events' into __init__.py

### LINKED ISSUE #305: Make the events module available in __init__.py
This is a quality of life thing. Currently, if I write code like this:

    import ppb

    class Player(ppb.BaseSprite):
        def on_update(self, update: ppb.events.Update, signal):
            pass

PyCharm complains "Cannot find reference 'events' in '__init__.py'"

We can fix this by importing it explicitly in the init file.

BubbsTheSupreme:
This resolves #305. I used VS Code to test.


BubbsTheSupreme:
If you don't mind me asking, what is `__all__` referring to? Is it a file or multiple files? I'm new to this project (and contributing to open source) so I'm not quite sure.

astronouth7303:
https://github.com/ppb/pursuedpybear/pull/334/files#diff-4f4e1b10744df4e400dcb41147d6df6dR11-R16

astronouth7303:
Also, please look at CI.

ppb/__init__.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import events
```
pathunstrom:
```suggestion
```

ppb/__init__.py
```diff
@@ -1,4 +1,5 @@
...
 from typing import Callable
```
pathunstrom:
```suggestion

from ppb import events
```

pathunstrom:
@BubbsTheSupreme `__all__` defines exported names for usage in `from ppb import *`. You should add to the shortcuts line the string `"events"` to get what @astronouth7303 is asking for.

ppb/__init__.py
BubbsTheSupreme: Should I change the `import events` to `from ppb import events` ?

BubbsTheSupreme: I tried this and it passed the 9 tests!

Issue #249, opened 2019-05-06
Targeted Events


pathunstrom:
We want to add events that get published to specific targets.

There's a discussion to be had on whether targeted events can _only_ be seen by a single target, or if we want to support multiple targets.

This change will _absolutely_ require a change to the engine itself, so is not recommended for new contributors.

jamesBrosnahan:
Can lambdas/method references be used with the current event system to make event targets similar to C# delegates or is this different from that type of functionality?

astronouth7303:
The exact mechanics are:

1. The engine produces a pile of potential event receivers
2. The `__event__()` is called on each receiver
3. `__event__()` handles dispatching to the local `do_*()` callable and also handles logging, error handling, etc.

We have no opinion about what kind of callable it is, but there is not currently way to add bare functions. The recommended solution would be: Wrap it in a system and add it to the engine. (#263)

astronouth7303:
The targeted thing is meant to handle cases where you want to use the event dispatch mechanism, but you don't want it broadcast to every active object in the game tree. This might be for: The renderer informing cameras about their geometry (in multi-camera contexts), collision systems only telling the objects that collided about their collision, and possibly other uses.

astronouth7303:
Note that this is different from #252, where the specific target is not necessarily known, but we (as engine and game designers) are fairly certain it should be handled by one object. (This is really an abuse of the event system in lieu of a DI or IOC framework.)

pathunstrom:
Okay, so this conversation gave me some inspiration, so going to forward a proposal:

Assuming that our current model for publishing looks like this (this is how it currently works, just not explicitly):

1. Pop the next event off the queue.
2. Call the registered extensions to extend or enrich the event.
3. Publish to the targets of the event (currently, the entire tree)

The suggestion is let's add `__targets__` as part of the events API. `__targets__` will by necessity be an iterable. We then modify step three above:

3. Publish to the targets of the event: If `__targets__` exists and is not empty, we only call `__event__` on its members, otherwise, we walk the entire tree as normal.

This gives us a mechanism by which #252 can be done, and keeps exceptions in the engine at a minimum. It also allows subsystems to use the extension plumbing to define it's API event simply by modifying __targets__ on those events.

PR #338, opened 2019-08-07
Colliders and ColliderSystem feature.

pathunstrom:
Adding simple circle and square collision mixins and a system that does the checks for you after an Update event fires and the queue clears.

### DIFF
--- a/ppb/features/collision.py
+++ b/ppb/features/collision.py
@@ -1,0 +1,108 @@
+"""
+A collision subsystem with various colliders available.
+"""
+from itertools import product
+
+from ppb.sprites import BaseSprite
+from ppb.systemslib import System
+
+
+class CollidedWithMixin:
+    """
+    Provides a mechanism for CollisionCheckerSystem to inform sprites when
+    they collide with something.
+
+    This is a mixin added to CanCollideCircleMixin and
+    CanCollideSquareMixin. You do not need to to add it to your sprites if
+    you are using one of them.
+    """
+
+    def collided_with(self, other):
+        """
+        Another sprite has collided with this sprite.
+
+        This is meant to be overridden in your sprites. If a sprite has a
+        collider mixin, but doesn't respond to those collisions, you can
+        ignore this.
+        """
+
+
+class CanCollideCircleMixin(BaseSprite, CollidedWithMixin):
+    """
+    Defines a circular region collider and provides a method to check for
+    collision.
+    """
+    radius: float = 1
+
+    def collides_with(self, other: 'CanCollideCircleMixin'):
+        """
+        Check if this sprite collides with another sprite.
+
+        Both sprites must be CanCollideCircleMixin sprites.
+        """
+        distance = (self.position - other.position).length
+        collide_distance = self.radius + other.radius
+        return distance <= collide_distance
+
+
+class CanCollideSquareMixin(BaseSprite, CollidedWithMixin):
+    """
+    Uses the defined square on a sprite to provid a method to check for
+    collision.
+    """
+
+    def collides_with(self, other: 'CanCollideSquareMixin'):
+        """
+        Check if this sprite collides with another sprite.
+
+        Both sprites must be CanCollideSquareMixin sprites.
+        """
+        furthest_left = min(self.left, other.left)
+        furthest_right = max(self.right, other.right)
+        furthest_up = max(self.top, other.top)
+        furthest_down = min(self.bottom, other.bottom)
+        collide_distance = self.size + other.size
+        return all([furthest_right - furthest_left <= collide_distance,
+                    furthest_up - furthest_down <= collide_distance])
+
+
+class CollisionCheckerSystem(System):
+    """
+    A subsystem to check for collisions after the update step of the system.
+
+    To use, add CollisionCheckerSystem to the list of systems in ppb.run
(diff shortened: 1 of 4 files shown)


ppb/features/collision.py
```diff
@@ -0,0 +1,108 @@
...
+    def on_idle(self, event, signal):
+        if not self.primed:
+            return
+        for s1, s2 in product(event.scene, event.scene):
+            if s1 is s2:
+                continue
+            try:
+                if s1.collides_with(s2):
+                    s1.collided_with(s2)
+                    s2.collided_with(s1)
```
astronouth7303: This line will mean that `.collided_with()` will be called twice.

pathunstrom: . . . yes. From two different objects, who are likely to have two different collided_with methods.

astronouth7303: No, I mean it'll detect twice.

First as s1 as Spam and s2 as Foo, and then again as s1 as Foo and s2 as Spam.

pathunstrom:
I mean, we've discussed: I want proper circle collision or proper AABB collision over point-in-region, which is what you keep forwarding. That's great for your games, but I must have one of these two for the games I make.

That said, this system doesn't make decisions on the kind of collision you're going to use only that you can call collides_with on either sprite with the other sprite and get a sensible result. So if you want fall-through collision (IE this is a circle collider, that's a square collider, do a thing that makes sense with those) only requires writing your own mixin with no changes to the (completely optional) subsystem.

I could add a point-in-region alternative, but it creates a symmetry problem: You get a different result if you do `s1.collides_with(s2)` versus `s2.collides_with(s1)` if the sprites are different sizes.

Arguably, this is _also_ a thing that really needs #249, using a defined method (`collided_with`) is just working around that problem. (I'd definitely prefer an `on_collided()` event that goes only to the sprites in question.

ppb/features/collision.py
pathunstrom: No.

`s1.collides_with(s2)` == check for collision.

`s1.collided_with(s2)` == s1 responds to a collision detected with s2.
`s2.collided_with(s1)` == s2 responds to a collision detected with s1.

ppb/features/collision.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    primed = False
+
+    def on_update(self, event, signal):
+        self.primed = True
+
+    def on_idle(self, event, signal):
+        if not self.primed:
+            return
+        for s1, s2 in product(event.scene, event.scene):
```
pathunstrom: Leaving a note for myself, I probably want combinations(event.scene, 2) and not product. Should add a test case to prove it.

ppb/features/collision.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Provides a mechanism for CollisionCheckerSystem to inform sprites when
+    they collide with something.
+
+    This is a mixin added to CanCollideCircleMixin and
+    CanCollideSquareMixin. You do not need to to add it to your sprites if
+    you are using one of them.
+    """
+
+    def collided_with(self, other):
```
pathunstrom: This should probably look like an event handler: `on_collided_with(self, event, signal)`.

The subsystem is going to skip the event plumbing because it's not strictly necessary, but it'll make it more clear what's going on.

ppb/features/collision.py
pathunstrom: Just making sure, you mean the product versus combination problem (from the top this block?)

astronouth7303: Yup. Product is probably what you want for asymmetric collisions, though.

astronouth7303:
That's... not my point.

Like, we both agree that the collision feature should allow sprites a way to declare their physical region.

We've had two implementations of this so far:
* Mine from mutant-games, which is point-in-region
* This PR, which is region-on-region.

The caveat of mine is that two regions will significantly overlap before a collision is detected.

The caveat of yours is that a sprite applies its own concept of region to other sprites, regardless of how that other sprite is declared. (If you have an AABB sprite `S` and a round sprite `C`, if you ask `S` if it collides with `C`, `S` will treat `C` as an AABB even though `C` is declared as circular.)

(To be clear, I will not say that my point-in-region system is right for all or even most games. It is just Very Simple and acceptable for my use case.)

Both these systems are asymetric--you can situations with S1 and S2 where S1 says there is a collision but S2 does not.

My argument is that asymmetry is an undesirable property in a general-purpose collision system.

My proposal to resolving this is to move the collision detection code out of the sprites and have it live independently. We can then define functions for each pairing of Circle, AABB, Point, and whatever else.

pathunstrom:
See documentation included: This assumes you're using the same kind of collision for all of your sprites. That's a decision I feel comfortable making. If you want to do the step through sort of collision (least accurate to most accurate, EG: circle, followed by AABB, followed by mask) that's a thing that's still possible with this system, it just requires overriding the basic collision method.

To change to point-in-region, you just have to do the reverse check as part of the `collides_with` method, and it will all just work.

astronouth7303:
That means that the choice of geometry is global, not on the sprite level. IMO, the API should reflect this.

On the subject of "We shouldn't have to apologize for our API": "Why can't I have round and square sprites?" (Because you know that question is going to come up.)

PR #357, opened 2019-08-24
Refactors sprites.BaseSprite

### LINKED ISSUE #347: Break up ppb.BaseSprite
BaseSprite started as a simple stub that was intended for subclassing and combining mixins. As time has gone on, it's collected a lot of weight. A while back @astronouth7303 suggested breaking this out as a genuine Base and I'm agreeing whole heartedly on it.

At this point, the things I consider core to BaseSprite are three things:

1. Event handlers
2. Position
3. The wonderful [`__init__`](https://github.com/ppb/pursuedpybear/blob/master/ppb/sprites.py#L222-L238) function.

Everything else we've built (even the [Sides API](https://github.com/ppb/pursuedpybear/blob/master/ppb/sprites.py#L22-L165) that I built early on and is [currently on the chopping block](https://github.com/ppb/pursuedpybear/issues/339)) are secondary and should be moved to mixins.

Finally, the import into `ppb.__init__` should provide a ppb.Sprite that is the thing users should subclass if they want all of the features.

This is a huge amount of work to be a single issue, so the immediate details will become their own issues, linked in the checklist below. This ticket is done when the dependent tickets are done.

- [ ] [SquareShapeMixin](https://github.com/ppb/pursuedpybear/issues/352)
- [ ] [RenderableMixin](https://github.com/ppb/pursuedpybear/issues/353) 
- [ ] [ppb.Sprite](https://github.com/ppb/pursuedpybear/issues/354)
- [ ] [ppb.BaseSprite deprecation](https://github.com/ppb/pursuedpybear/issues/355)

### LINKED ISSUE #352: Add a SquareShapeMixin
This is a subissue of #347 

We need to pull out the Sides API  and `size` attribute to a new mixin: `SquareShapeMixin`.

Eventually, this mixin can inherit from `BaseSprite` (so it would not need to define its own `position`)  but until 347 is finished, it's going to need to stand alone.

Steps to complete:

- [ ] Define new class `SquareShapeMixin`
- [ ] Add class attribute 'size' with type `float` default to 1.
- [ ] Add class attribute 'position' with type `ppb_vector.Vector` default to `ppb_vector.Vector(0, 0)`
- [ ] Add the sides API to `SquareShapeMixin`
- [ ] Remove sides API from `BaseSprite`
- [ ] Remove size from `BaseSprite`
- [ ] Include `SquareShapeMixin` in `BaseSprite` parent classes.

### LINKED ISSUE #353: Make RenderableMixin
This is a subissue of #347 

We need to pull out the `__image__` method, `image` attribute and size to a new mixin: `RenderableMixin`.

Eventually, this mixin can inherit from BaseSprite (so it would not need to define its own position) but until 347 is finished, it's going to need to stand alone. Additionally, because this mixin interacts with the renderer, you'll need to keep a stub `__image__` method that returns `DoNotRender` until the renderers API is updated.

To complete:

- [ ] Create class `RenderableMixin`
- [ ] Add a `size` class attribute with the type `int` and a default of 1 to `RenderableMixin`.
- [ ] Add a `position` class attribute with the type `ppb_vector.Vector` and a default of `Vector(0, 0)` to `RenderableMixin`.
- [ ] Add an `image` class attribute with the type `Union[Flag, Asset]` and a default value of `None` to `RenderableMixin`.
- [ ] Add an `__image__` method that is a copy of the one on `BaseSprite` to `RenderableMixin`.
- [ ] Remove `image` class attribute from `BaseSprite`.
- [ ] Change `BaseSprite.__image__` to return `DoNotRender`

### LINKED ISSUE #354: Add a ppb.Sprite
`ppb.Sprite` will be the new default import for sprites.

It will be a subclass of `BaseSprite`, and all of our mixins: `Rotatable`, `SquareShapeMixin`, `RenderableMixin`.

This definition will live in `ppb.__init__`.

Depends on #352 #353 

Subissue of #347

### LINKED ISSUE #355: Deprecate ppb.BaseSprite in favor of ppb.Sprite
Like many breaking changes in `ppb` before 1.0, we want to give a minimum of 1 release of deprecation warning. To do so, we're going to need to make a `BaseSprite` stub in `ppb.__init__` that warns of deprecation, with instructions on how to use the new `BaseSprite` (by importing the `sprites` module) or to use the new `ppb.Sprite`

Subissue of #347 

Depends on #354

pathunstrom:
I was just going to handle the rename, but decided to do all of this together with an okay from @astronouth7303. Not ready for merge, going to clean up the docs around Sprites and add tests for ppb.sprites.Sprite and ppb.BaseSprite. Additionally, would like to cleave out the various bits from the mixins into their own tests.

Resolves #347 
Resolves #352 
Resolves #353 
Resolves #354 
Resolves #355

### DIFF
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -6,25 +6,26 @@ from ppb import events
 from ppb_vector import Vector
 from ppb.engine import GameEngine
 from ppb.scenes import BaseScene
-from ppb.sprites import BaseSprite as OldBaseSprite
+from ppb.sprites import Sprite
 from ppb.systems import Image
 from ppb.systems import Sound
 
 __all__ = (
     # Shortcuts
-    'Vector', 'BaseScene', 'BaseSprite', 'Image', 'Sound', 'events',
+    'Vector', 'BaseScene', 'BaseSprite', 'Image', 'Sprite', 'Sound',
+    'events',
     # Local stuff
     'run', 'make_engine',
 )
 
 
-class BaseSprite(OldBaseSprite):
+class BaseSprite(Sprite):
     """
     A stub that raises a deprecation warning when a user uses
     ``ppb.BaseSprite.``
     """
     __warning = """Using ppb.BaseSprite is deprecated.
-    
+
     You probably want ppb.Sprite. If you're wanting to use BaseSprite and
     mixins to change what features your sprites have, import
     ppb.sprites.BaseSprite.
@@ -32,12 +33,6 @@ class BaseSprite(OldBaseSprite):
 
     def __init__(self, *args, **kwargs):
         warnings.warn(self.__warning, DeprecationWarning)
-
-
-class Sprite(OldBaseSprite):
-    """
-    A stub pointing to current functionality under the new name.
-    """
 
 
 def _make_kwargs(setup, title, engine_opts):
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -31,8 +31,9 @@ class BaseSprite(Sprite):
     ppb.sprites.BaseSprite.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, **kwargs):
         warnings.warn(self.__warning, DeprecationWarning)
+        super().__init__(**kwargs)
 
 
 def _make_kwargs(setup, title, engine_opts):
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -1,11 +1,12 @@
 import logging
+import warnings
 from typing import Callable
 
 from ppb import events
 from ppb_vector import Vector
 from ppb.engine import GameEngine
 from ppb.scenes import BaseScene
-from ppb.sprites import BaseSprite
+from ppb.sprites import BaseSprite as OldBaseSprite
 from ppb.systems import Image
 from ppb.systems import Sound
 
@@ -15,6 +16,28 @@ __all__ = (
     # Local stuff
     'run', 'make_engine',
 )
+
+
+class BaseSprite(OldBaseSprite):
+    """
+    A stub that raises a deprecation warning when a user uses
+    ``ppb.BaseSprite.``
+    """
+    __warning = """Using ppb.BaseSprite is deprecated.
+    
+    You probably want ppb.Sprite. If you're wanting to use BaseSprite and
+    mixins to change what features your sprites have, import
+    ppb.sprites.BaseSprite.
+    """
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn(self.__warning, DeprecationWarning)
+
+
+class Sprite(OldBaseSprite):
+    """
+    A stub pointing to current functionality under the new name.
+    """
 
 
 def _make_kwargs(setup, title, engine_opts):
--- a/ppb/sprites.py
+++ b/ppb/sprites.py
@@ -19,6 +19,46 @@ error_message = "'{klass}' object does not have attribute '{attribute}'"
 side_attribute_error_message = error_message.format
 
 
+class RotatableMixin:
+    """
+    A simple rotation mixin. Can be included with sprites.
+    """
+    _rotation = 0
+    # This is necessary to make facing do the thing while also being adjustable.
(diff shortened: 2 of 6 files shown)


ppb/__init__.py
```diff
@@ -1,22 +1,41 @@
...
+    ``ppb.BaseSprite.``
+    """
+    __warning = """Using ppb.BaseSprite is deprecated.
+
+    You probably want ppb.Sprite. If you're wanting to use BaseSprite and
+    mixins to change what features your sprites have, import
+    ppb.sprites.BaseSprite.
+    """
+
+    def __init__(self, **kwargs):
```
astronouth7303: Let's move this warning to `__init_subclass__()`

ppb/sprites.py
```diff
@@ -19,6 +26,113 @@
...
+class BaseSprite(EventMixin):
+    """
+    The base Sprite class. All sprites should inherit from this (directly or
+    indirectly).
+
+    The things that define a BaseSprite:
+
+    * The __event__ protocol (see ppb.eventlib.EventMixin)
+    * A position vector
+    * A layer
```
astronouth7303: Layer only makes sense on renderables

ppb/sprites.py
```diff
@@ -19,6 +26,113 @@
...
+            if k == 'position':
+                v = Vector(v)
+            setattr(self, k, v)
+
+
+class RenderableMixin:
+    """
+    A class implementing the API expected by ppb.systems.renderer.Renderer.
+
+    You should include RenderableMixin after BaseSprite in your parent class
```
astronouth7303: `RenderableMixin` should come before `BaseSprite`, assuming that `BaseSprite` is the conceptual parent to `RenderableMixin`.

ppb/sprites.py
```diff
@@ -165,74 +279,20 @@ def _attribute_gate(self, attribute, bad_sides):
...
-    def rotate(self, degrees):
-        """
-        Rotate the sprite by a given angle (in degrees).
-        """
-        self.rotation += degrees
-
+    A mixin that applies square shapes to sprites.
 
-class BaseSprite(EventMixin, Rotatable):
+    You should include SquareShapeMixin after ppb.sprites.BaseSprite in
```
astronouth7303: Again, before.

ppb/systems/renderer.py
```diff
@@ -15,6 +15,8 @@
 DEFAULT_RESOLUTION = 800, 600
 
 
+# TODO: Move Image out of the renderer so sprites can type hint
+#  appropriately.
```
astronouth7303: I've been type-hinting with `Asset` to avoid this, but yeah, we have a circular dependency problem.

ppb/__init__.py
pathunstrom: I'd worry about that since you can instantiate BaseSprite directly to build with.

ppb/sprites.py
pathunstrom: Not really.

`Scene.sprite_layers()` exists whether you have a renderer or not.

ppb/__init__.py
astronouth7303: Oh fair.

ppb/sprites.py
astronouth7303: "Renderable" means "can be rendered", as in, sprites with rendering support.

I don't think a "layer" makes sense on Sprites that don't have any rendering support (don't have `RenderableMixin`).

pathunstrom: There's more use cases for layering besides rendering. A simple use case is using layers to determine collisions (as in, only things on the same layer can collide).

pathunstrom: In general: The `Renderer` doesn't actually care if things have layer since it leans on `Scene.sprite_layers` for render order. The `Scene` does care, because it's the one that implements iteration in layer order. `Scene` doesn't care if things are `Renderable`, so layer needs to either go on `BaseSprite` or a separate mixin, and I . . . am cautious about making one attribute mixins.

astronouth7303: :thinking: 

I'm skeptical about just making up use cases that abuse semantics, but I'll call the argument "good enough".

Issue #349, opened 2019-08-24
logging.warn -> logging.warning


pathunstrom:
`Logger.warn` is deprecated, we should convert all logging usages over to `Logger.warning`

Issue #358, opened 2019-08-24
Bug in Camera.translate_to_frame


pathunstrom:
Hypothesis found a bug in the Camera. Relevant test failure:

```__________________________________________________________________ test_transfromation_movement __________________________________________________________________

    @given(
>       vp_width=st.integers(min_value=1),
        vp_height=st.integers(min_value=1),
        pixel_ratio=st.floats(min_value=1, max_value=1e5, allow_nan=False, allow_infinity=False),
        cam_pos=vectors(),
        point=vectors(),
        delta=vectors(),
    )
    def test_transfromation_movement(
        vp_width, vp_height, pixel_ratio, cam_pos, point, delta,
    ):

tests\test_camera.py:131:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

vp_width = 1, vp_height = 1, pixel_ratio = 1.0, cam_pos = Vector(0.0, 7.067388259113906e+72), point = Vector(0.0, 0.0), delta = Vector(0.0, 2.2204460492503134e+59)


    @given(
        vp_width=st.integers(min_value=1),
        vp_height=st.integers(min_value=1),
        pixel_ratio=st.floats(min_value=1, max_value=1e5, allow_nan=False, allow_infinity=False),
        cam_pos=vectors(),
        point=vectors(),
        delta=vectors(),
    )
    def test_transfromation_movement(
        vp_width, vp_height, pixel_ratio, cam_pos, point, delta,
    ):
        cam = Camera(
            viewport=(0, 0, vp_width, vp_height),
            pixel_ratio=pixel_ratio,
        )
        cam.position = cam_pos
    
        assume(delta.length != 0)
    
        point_moved = point + delta
    
        assume(point_moved != point)  # This will happen when delta is too small to make an effect
    
        note(f"point moved: {point_moved}")
    
        diff = cam.translate_to_frame(point_moved) - cam.translate_to_frame(point)
        note(f"frame diff: {diff}")
>       assert isclose(delta.length, pixel_ratio * diff.length, rel_tol=1e-4)
E       assert False
[...]

PR #359, opened 2019-09-03
Logger.warn is converted to Logger.warning

arulprabakaran:
Logger.warn is converted to Logger.warning

Fixes: #349


astronouth7303:
The failed test is #358.

pathunstrom:
@arulprabakaran Thank you for your PR. We'll be getting to it soon, but as @astronouth7303 mentioned in #360 we encourage contributors to include their name in the same PR as their first contribution. If you'd rather, we can work on #360?

PR #363, opened 2019-09-07
Adds Primitive Assets.

### LINKED ISSUE #337: Drawing Primitives Assets
The start of a massive change to the way PPB deals with rendering:

Instead of `Sprite.image = None` rendering a square, I'd like to provide an explicit API both for rendering the drawing primitives and defining their color, if they are an outline or a fill, and other such features.

Examples:

```
class SquareSprite(ppb.BaseSprite):
    image = ppb.primitives.Square(color=(255, 50, 50, 1), outline=None)
```

This square will be red, full opacity, and filled.

```
class CircleSprite(ppb.BaseSprite):
    image = ppb.primitives.Circle(color=(40, 255, 50, 0.5), outline=5)
```
This circle will be green, half opacity, and will have an outline of approximately 5 pixels when size 1.

```
class TriangleSprite(ppb.BaseSprite):
    image = ppb.primitives.Triangle(color=(0, 0, 0), outline=None)
```
A black triangle, filled.

Each shape is going to be a pseudo Asset.

pathunstrom:
Relies on #361 (since rendering is broken without it.)

Resolves #337

We don't hold on to any information in this class, the color is used and discarded immediately, encouraging the same usage as other asset classes.

### DIFF
--- a/ppb/assets.py
+++ b/ppb/assets.py
@@ -20,11 +20,11 @@ def _create_surface(color):
     """
     Creates a surface for assets and sets the color key.
     """
-    asset = Surface(DEFAULT_SPRITE_RESOLUTION)
+    surface = Surface(DEFAULT_SPRITE_RESOLUTION)
     color_key = BLACK if color != BLACK else MAGENTA
-    asset.set_colorkey(color_key)
-    asset.fill(color_key)
-    return asset
+    surface.set_colorkey(color_key)
+    surface.fill(color_key)
+    return surface
 
 
 class Shape(AbstractAsset):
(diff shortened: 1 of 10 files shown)


ppb/assets.py
```diff
@@ -1,209 +1,80 @@
...
-        """
-
-    def is_loaded(self):
-        """
-        Returns if the data is ready now or if :py:meth:`load()` will block.
-        """
-        return True
+    Creates a surface for assets and sets the color key.
+    """
+    asset = Surface(DEFAULT_SPRITE_RESOLUTION)
```
astronouth7303: Not a fan of this name.

ppb/assets.py
```diff
@@ -1,209 +1,80 @@
...
-            import traceback
-            traceback.print_exc()
-            # Save unhandled exceptions to be raised in the main thread
-            self._raise_error = exc
-        finally:
-            # This always needs to happen so the main thread isn't just blocked
-            self._finished.set()
-
-    def background_parse(self, data: bytes):
+    def modify_asset(self, color: Tuple[int, int, int]) -> None:
```
astronouth7303: nor this one.

ppb/assets.py
pathunstrom: The default sprite resolution or asset?

astronouth7303: `asset`

pathunstrom:
Yeah, I'll take that waggle. To be fair, this should probably have been patterned on eventlib and systemslib from the start. :-P

pathunstrom:
bors retry

PR #368, opened 2019-09-23
Sound Docs

astronouth7303:
Some basic reference docs for sound.


pathunstrom:
Comment for us to handle later, but I realize we don't have a sample of instantiating a Sound or Image anywhere in the docs. We should probably do that at some point.

PR #376, opened 2019-12-09
Adds RectangleShapeMixin

### LINKED ISSUE #339: Sides refactor
Ok, a few things I'd like to do with sides:

* Replace `.left.top` etc with `.left_top` so that the whole thing is significantly less magical
* Make the sides API a general mixin, so its usable by camera and bounded scenes
* Use this as part of the camera-is-not-sprite rebuild

### LINKED ISSUE #372: Non-square sprites
Break the assumption that all sprites are square.

* Change `size` to `width`/`height`
* Figure out the best way to tie together the asset dimensions and declared sprite dimensions
* How to handle mismatched aspect ratios?

pathunstrom:
Includes moving away from the Sides interface on this new sprite type.

- [x] Sprite sides setters (top, left, right, bottom) and tests
- [x] Rewrite hypothesis tests to cover new sprites
- [x] SquareSpriteMixin (hook into tests)
- [x] RectangleSprite (concrete sprite that is Sprite, but doesn't have the size interface.)
- [x] Sprite to use SquareSpriteMixin
- [x] Camera to use RectangleShapeMixin
- [x] `Sprite.side_center` or `Sprite.side_middle`

Closes #372 
Closes #339


pathunstrom:
Yup. Mostly threw the properties at it to get the shape of the thing.

ppb/sprites.py
```diff
@@ -141,228 +133,127 @@ def rotate(self, degrees):
...
-        TOP: ('y', 1),
-        BOTTOM: ('y', -1)
-    }
+    A Mixin that provides a rectangular area to sprites.
 
-    def __init__(self, parent: 'SquareShapeMixin', side: str):
-        self.side = side
-        self.parent = parent
+    You should include RectangleShapeMixin before your BaseSprite in your
+    parent classes.
```
astronouth7303, line 141: Why before? Is it mandatory that `RectangleShapeMixin` happen before `BaseSprite`?

ppb/sprites.py
```diff
@@ -141,228 +133,127 @@ def rotate(self, degrees):
...
-        return f"Side({self.parent!r}, {self.side!r})"
+    @property
+    def left(self) -> float:
+        return self.position.x - self.width / 2
 
-    def __str__(self):
-        return str(float(self))
+    @left.setter
+    def left(self, value: Union[float, int]):
+        self.position = Vector(value + (self.width * 0.5), self.position.y)
```
astronouth7303, line 160: Why `* 0.5` instead of `/ 2`?

ppb/sprites.py
```diff
@@ -141,228 +133,127 @@ def rotate(self, degrees):
...
-    def right(self) -> Side:
-        """
-        The right side
-        """
-        return Side(self, RIGHT)
-
-    @right.setter
-    def right(self, value):
-        self.position = Vector(value - self._offset_value, self.position.y)
+class SquareShapeMixin(RectangleShapeMixin):
```
astronouth7303, line 239: Docstring

ppb/sprites.py
```diff
@@ -378,3 +269,16 @@ class Sprite(SquareShapeMixin, RenderableMixin, RotatableMixin, BaseSprite):
...
+class RectangleSprite(RectangleShapeMixin, RenderableMixin, RotatableMixin, BaseSprite):
+    """
+    A rectangle sprite.
+
+    Sprite includes:
+
+    * BaseSprite
+    * RectangleShapeMixin
+    * RenderableMixin
+    *RotatableMixin
```
astronouth7303, line 283: Missing space

ppb/sprites.py
pathunstrom: This was a thing we discussed before? You had said that mixins should come before the base class as a general rule?

ppb/sprites.py
pathunstrom: No particular reason, willing to change.

ppb/sprites.py
astronouth7303: Yeah, but documenting it this way kinda implies "`RectangleShapeMixin` must be used with `BaseSprite`"? Which is untrue, and if it was true it should inherit from `BaseSprite`?

pathunstrom: Would "You should include RectangleShapeMixin before you base class in your class definitions"?

astronouth7303: I don't think it actually matters in this context? Because it doesn't override anything (just assumes certain things are defined somewhere in the MRO), I don't think it matters?

PR #379, opened 2020-02-19
Update Mac CI

astronouth7303:
Mac CI broke. Also, there's new osx versions.


pathunstrom:
CI passes, looks good to me.

bors r+

PR #380, opened 2020-02-19
Mechanized Release Process

astronouth7303:
Implements a mechanized release process. Under this, the release process would be:

1. Go to GitHub Release pages
2. Look at the diff since last version, compile changelog
3. On the GitHub site, create a tag/release with the changelog and any other release notes
4. Cirrus CI is responsible for building and uploading to PyPI and GitHub

Additionally, dev releases are uploaded to PyPI test every time `master` is updated.

Version numbers are no longer explicitly listed in the repo, instead being calculated based on the git tags.

This code is a pretty direct copy from https://github.com/gqlmod/gqlmod (which is in part based on code that @duckinator wrote).

@nbraud originally proposed this for PPB.


astronouth7303:
Note that I don't really expect this to be merged. It's more along the lines of "demo then discuss".

astronouth7303:
Ok, I'm going to leave the shuffling of artifacts around until we decide to actually do this.

pathunstrom:
> Additionally, dev releases are uploaded to PyPI test every time master is updated.

Alright this is pretty much the thing that settles my doubts. It gives us a live preview of the build artifacts before we cut a release. If you and @nbraud are happy with this, I'm happy with it.

astronouth7303:
Ok, CI is green.

Just need to add the API tokens, and this should be good.

astronouth7303:
NVM. I can't log in as the bot because of Security:tm:, so the credentials will have to wait.

astronouth7303:
Tokens are in place.

Tested uploads.

I think this is good.

pathunstrom:
Off we go into the new world.

bors r+

astronouth7303:
bors retry

astronouth7303:
bors retry

PR #387, opened 2020-03-15
Move to PySDL

### LINKED ISSUE #197: Engine Tracking Issue
Ok, so: We've officially outgrown PyGame. Basic work with pyglet show a massive performance improvement.

For due dilligence, we should also test direct SDL2 bindings. Are there other libraries we should check?

Questions we should be asking:

* What platforms are supported?
* What is the installation story for game developers using ppb?
* What is the performance of it? relative to pygame? other candidates?
* What work does this save us?
* What additional work does this add?
* How does this affect future work: Sound? Animation? Packaging? Canvas Revolution (including text)?
* (Jamie's special interest: How does this interact with asyncio?)

### LINKED ISSUE #377: Migrate away from PyGame
PyGame has been broken on Python 3.8 since 3.8 started its release process (we removed 3.8 support [9mo ago](https://github.com/ppb/pursuedpybear/commit/3eb44f333279b0a49655adf82ba48f532feb90a8)). The PyGame project has no plans to fix this:

[PPB](https://twitter.com/pursuedpybear/status/1229454583610716160):
> Hey @pygame_org! Any chance of a Py3.8-compatible release? This has become a serious problem for us.

[PyGame](https://twitter.com/pygame_org/status/1229822479515955201):
> No.

The background is that PyGame has abandoned any PyGame 1 development or maintenance and is devoting all of their effort to PyGame 2. However, there is currently no indications if or when PyGame 2 will have a release of any kind.

PyGame is causing us to break our own support goals (always the latest Python, tracking the PSF's own support timelines), and they have no plans to fix this.

This is untennable for PPB, so we must consider our options.

1. Get PyGame to fix it (This would require extraordinary measures)
2. Fix PyGame for them (we'd be investing in an external project for an indeterminate amount of time to limp it along until its actual maintainers decides to do their job)
3. Change PPB's backend engine (proven extremely doable by the [pyglet spike](https://github.com/ppb/pursuedpybear/pull/169), there's just no slam-dunk backends available)

My assessment is that Option 3 will get us the most long-term value. Before discussing possible backends, a quick run through criteria:

* Stable on Linux, Windows, and Mac
* Supports Python versions roughly following PSF's Python lifetime schedules
* Installable through the Python ecosystem with no additional steps
* Must not compromise our own goals (eg, force us to apologize about gotchas)

The backend options we're aware of are:
* Pyglet: A pure-python system with its own platform code for Windows, Mac, and Linux/X11
* Thin SDL bindings: There are several projects in this category, some are C, some are ctypes
* Write our own platform code: While this is technically an option, none of us want to do it and we lack expertise to execute on 3-platform support

Pyglet's major problem is that there are crashing bugs on mac. (https://github.com/pyglet/pyglet/issues/5 https://bitbucket.org/pyglet/pyglet/issues/199/attempting-to-resize-or-close-pyglet https://bitbucket.org/pyglet/pyglet/issues/229/closing-window-causes-a-crash-on-mac-os) However, I would guess that after these problems are fixed, we would need to invest little-to-none in Pyglet for it to continue to work.

The SDL bindings currently have primarily shipping problems--they don't ship 3 platforms of binary wheels with SDL bundled. This is something that can be fixed with the application of CI/CD. We may even find it beneficial to adopt or fork one of the projects. I expect on-going upkeep to be low, just keeping up with CPython and SDL releases.

What direction we go in is a high-level strategic decision.

astronouth7303:
Completely replace PyGame with PySDL, per #377.

This should cleanly and completely replace PPB's use of PyGame with PySDL.

* [x] Rendering
* [x] Sound
* [x] Input
* [x] Abstract shapes

Closes #377

Closes #197


astronouth7303:
Because linux will need package installers until we invest in pysdl2-dlls, here's a running list of needed packages:

Debian: `libsdl2-mixer-2.0-0`

astronouth7303:
Renderer does need an optimization pass.

I think `ppb_mutant.index` got _worse_. It probably has to do with the hundreds of `SDL_CreateTextureFromSurface()` made every frame.

pathunstrom:
> Renderer does need an optimization pass.
> 
> I think `ppb_mutant.index` got _worse_. It probably has to do with the hundreds of `SDL_CreateTextureFromSurface()` made every frame.

I'd create the textures only the one time when we marshal the load, that's definitely a "should be outside the renderer" thing.

astronouth7303:
SDL_Surfaces are created outside and independently of renderers.

SDL_Textures are bound to a specific `SDL_Renderer`.

When I get everything working again, I'll give it a second chance to memoize texture creation.

pathunstrom:
Consider storing them in a dict on the Assets? I assume an SDL Renderer has an id and can be used as a key?

astronouth7303:
Want to index against Surfaces not Assets because mutable proxy assets.

My previous attempts involved a `WeakKeyDictionary` POINTER(Surface) -> POINTER(Texture), but apparently pointers aren't hashable. I forgot I was working with pointers not direct structures, though, and set the whole thing aside to actually finish functionality.

pathunstrom:
So, the issue I see with that is that surfaces are _also_ mutable. One of us should test how they hash against each other.

astronouth7303:
Ok, so:

This PR works: Tests and vistests both pass. `mutant-games` functions (although parallax scenes are jumpy again (still?)). Shooter game seems to work (although sometimes it seems like an arrow gets stuck and it just slides off the side of the screen?) Examples are also working.

astronouth7303:
_oh_.

Both shootergame a targets are exhibiting the same weird key input bug.

astronouth7303:
Ok, yeah, i forgot to discard repeat keypresses. Fixed.

astronouth7303:
Aaaaaaaaand SDL is crashing on mac, and signs point to a bug in SDL itself.

Test script: https://gist.github.com/astronouth7303/4d232f6ea166801b06d276dc139cb107

Crashes when you use the touchpad to move the mouse on mac.

pathunstrom:
Tested this on mac and windows. Install hung the first time on my windows machine, but honestly that might have been the client hiccuping.

Only thing that didn't work was the sprite layering demo, and that was because those resources didn't load. We should go figure out _why_ but not part of this work since all the rest worked just fine.

bors r+

pathunstrom:
bors retry

pathunstrom:
bors retry

astronouth7303:
Ok, the layering is weird. It loaded just fine for me. Did you call it right? You have to call it while sitting in the root of the project.

astronouth7303:
bors retry

pathunstrom:
bors retry

pathunstrom:
bors retry

pathunstrom:
bors retry

astronouth7303:
bors retry

astronouth7303:
bors retry

astronouth7303:
bors retry

astronouth7303:
And now bors itself is failing. Merging via CLI.

PR #415, opened 2020-04-04
Pysdl2 render optimizations

ironfroggy:
This includes @astronouth7303's weakref cache for image textures in addition to a FPS limit fix to avoid floating point precision errors that were cause the accumulator to increase at a slower rate than expected.

### DIFF
--- a/ppb/systems/renderer.py
+++ b/ppb/systems/renderer.py
@@ -38,6 +38,7 @@ import ppb.assetlib as assets
 import ppb.events as events
 import ppb.flags as flags
 from ppb.systems._sdl_utils import SdlSubSystem, sdl_call, SdlError
+from ppb.systems._utils import ObjectSideData
 
 logger = logging.getLogger(__name__)
 
@@ -138,6 +139,8 @@ class Renderer(SdlSubSystem):
         self.target_frame_rate = target_frame_rate
         self.target_count = 1 / self.target_frame_rate
 
+        self._texture_cache = ObjectSideData()
+
     def __enter__(self):
         super().__enter__()
         img_call(IMG_Init, IMG_INIT_JPG | IMG_INIT_PNG | IMG_INIT_TIF)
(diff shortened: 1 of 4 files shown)


ppb/systems/renderer.py
```diff
@@ -134,9 +136,12 @@ def __init__(
         self.window = None
         self.window_title = window_title
         self.pixel_ratio = None
-        self.render_clock = 0
+        self.render_clock = time()
```
astronouth7303, line 139: Can you use the monotonic clock instead of the irl civil wall time? Unless there's a reason to not use monotonic.

ppb/systems/renderer.py
```diff
@@ -163,12 +168,12 @@ def __exit__(self, *exc):
         super().__exit__(*exc)
 
     def on_idle(self, idle_event: events.Idle, signal):
-        self.render_clock += idle_event.time_delta
-        if self.render_clock > self.target_count:
+        self.render_clock = time()
```
astronouth7303, line 171: This looks like it can be a local?

ironfroggy: ✔️

ppb/systems/renderer.py
ironfroggy: ✔️

ppb/systems/renderer.py
```diff
@@ -168,12 +167,12 @@ def __exit__(self, *exc):
         super().__exit__(*exc)
 
     def on_idle(self, idle_event: events.Idle, signal):
-        self.render_clock = time()
-        if self.render_clock >= self.target_clock:
+        t = time()
```
astronouth7303, line 170: This also needs to be `monotonic()`--especially for consistency!

astronouth7303:
Initial testing shows a <1% increase in FPS between this branch and `master`. (500 frames per run, each branch ran 20 times.)

I find this extremely surprising, so further investigation is necessary.

pathunstrom:
Reminder that the renderer is designed to aim for 30fps, we could try unlocking it but not sure that's a good idea. Maybe aim for 60fps and see if you get a bigger jump?

astronouth7303:
Ok, profiling:

Running my "stress test" (`ppb_mutant.index` for 500 frames) under profiling shows that the cumulative time of `prepare_resource()` (time spent under that function and all sub-functions) by 80%. The time spent idling (specifically, the `on_idle` of `inputs.py`) increased by 42%--but this isn't where all the time freed by this change went to; it's going to a variety of places.

astronouth7303:
And to follow up on Piper's suggestion, I ran crabdance again at 1000FPS:

* master: 170.6FPS
* this PR: 318.5FPS

Ok, so that's much better. We should be looking for other reasons why we can't actually hit our target framerate.

astronouth7303:
Oh, the obvious one: My total run time numbers are the runtime of the entire process, not just the average time between frames.

PR #416, opened 2020-04-05
Asset Discussion

astronouth7303:
Add a discussion about the asset system.

### DIFF
--- a/docs/discussion/assets.rst
+++ b/docs/discussion/assets.rst
@@ -1,3 +1,5 @@
+.. py:currentmodule:: ppb.assetlib
+
 The Asset System
 ================
 
@@ -9,7 +11,7 @@ To that end, we had several goals when we built the asset system:
 
 * Allow declaring resources at a fairly abstract level
 * Optimistically load resources in the background as soon as possible
-* Provide a barrier of abstraction between how data is loaded and the use of that data
+* Provide a layer of abstraction between how data is loaded and the use of that data
 
 As part of this, we also built the VFS library (:py:mod:`ppb.vfs`), which treats the Python module import system as a filesystem and allows loading data from it, to make it clear where and how resource files should be added to a project, and provide all the flexibility of the Python module system.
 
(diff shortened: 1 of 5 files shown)


docs/discussion/assets.rst
```diff
@@ -0,0 +1,27 @@
...
+* Optimistically load resources in the background as soon as possible
+* Provide a barrier of abstraction between how data is loaded and the use of that data
+
+As part of this, we also built the VFS library (:py:mod:`ppb.vfs`), which treats the Python module import system as a filesystem and allows loading data from it, to make it clear where and how resource files should be added to a project, and provide all the flexibility of the Python module system.
+
+
+Concepts
+--------
+
+Our of this, we define a few high-level concepts:
```
pathunstrom, line 20:
```suggestion
Out of this, we define a few high-level concepts:
```

I think this is what you mean?

astronouth7303:
Ok, @pathunstrom, that's the first bit of word spew. What needs expanding or explaining?

docs/discussion/assets.rst
```diff
@@ -0,0 +1,67 @@
...
+So how did we do this?
+
+A lot of the heavy lifting is provided by the :py:mod:`concurrent.futures` package from the standard library. On top of this, :py:class:`AssetLoadingSystem` and :py:class:`Asset` cooperate to implement background file reading. After the data is read, it is handed to the instance and processed into its final form.
+
+Effort is taken to deduplicate assets: If two places refer to the same asset, it is normalized to the same instance. This reduces both load times and memory usage.
+
+A minor wrinkle in this is that assets are defined before the engine starts. The asset system does not actually begin loading data until the engine and :py:class:`AssetLoadingSystem` are initialized. This means that there's no problems delivering events and asset implementations know that initialization has happened.
+
+
+Usage
```
pathunstrom, line 44: This is fine for now, but this is definitely a thing I think we should provide some cookbook entries for in the future. I think we even have an issue for that.

pathunstrom:
I think it resolves the problem: it explains the whys and the more tricky whats, the usage bit will be better served in either tutorials or cookbook entries, but those don't exist, so having it here is at least helpful in the short term. We can pull them out when those exist.

Issue #394, opened 2020-03-20
Channel Management


astronouth7303:
Under #387, if the game tries to play more sounds than their are channels, the game crashes.

We need to do more active channel management and handle this. (Either by increasing channels or not playing the sound.)

PR #422, opened 2020-04-15
Fixes error when playing more sounds than channels.

### LINKED ISSUE #420: Playing too many longer sounds causes and SDL channels error
If you play more than 8 sounds at once, which seems to be the default number of channels, there is a non-recoverable error. Maybe this case should fail silently, possibly with a warning rather than an error. Or, an error handling mechanism could let users catch this. Or, the ability to change the number of channels could help the user have more control over the limit.

Most importantly, imho, would be a less-aggressive default failure mode.

```
Traceback (most recent call last):
  File ".\main.py", line 547, in <module>
    resolution=(1080, 800),
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\__init__.py", line 71, in run
    eng.run()
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\engine.py", line 100, in run
    self.main_loop()
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\engine.py", line 111, in main_loop
    self.loop_once()
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\engine.py", line 121, in loop_once
    self.publish()
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\engine.py", line 154, in publish
    method(event, self.signal)
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\systems\sound.py", line 149, in on_play_sound
    _check_error=lambda rv: rv == -1
  File "c:\users\calvin\gamedev\pursuedpybear\ppb\systems\sound.py", line 48, in _call
    raise SdlMixerError(f"Error calling {func.__name__}: {err.decode('utf-8')}")
ppb.systems.sound.SdlMixerError: Error calling <lambda>: No free channels available
```

pathunstrom:
Includes an increase in the default number of channels available to the sound system.

Regarding #420

This isn't a fix so much as a temporary patch until we can fix it more completely later.

Closes #420 

#394 would be a more complete implementation

### DIFF
--- a/ppb/systems/sound.py
+++ b/ppb/systems/sound.py
@@ -21,6 +21,7 @@ from sdl2.sdlmixer import (
 
 from ppb import assetlib
 from ppb.systems._sdl_utils import SdlError, SdlSubSystem
+from ppb.utils import LoggingMixin
 
 __all__ = ('SoundController', 'Sound', 'SdlMixerError')
 
@@ -89,7 +90,7 @@ def _filler_channel_finished(channel):
     pass
 
 
-class SoundController(SdlSubSystem):
+class SoundController(SdlSubSystem, LoggingMixin):
     _finished_callback = None
 
     def __init__(self, **kw):
@@ -153,6 +154,7 @@ class SoundController(SdlSubSystem):
         except SdlMixerError as e:
             if not str(e).endswith("No free channels available"):
                 raise
+            self.logger.warn("Attempted to play sound, but there were no available channels.")
         else:
             self._currently_playing[channel] = sound  # Keep reference of playing asset
 
--- a/ppb/systems/sound.py
+++ b/ppb/systems/sound.py
@@ -123,6 +123,7 @@ class SoundController(SdlSubSystem):
             # not sure how much difference it makes.
             _check_error=lambda rv: rv == -1
         )
+        self.allocated_channels = 16
 
         # Register callback, keeping reference for later cleanup
         self._finished_callback = channel_finished(self._on_channel_finished)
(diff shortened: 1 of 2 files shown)


ppb/systems/sound.py
```diff
@@ -141,14 +142,18 @@ def on_play_sound(self, event, signal):
...
-        self._currently_playing[channel] = sound  # Keep reference of playing asset
+        try:
+            channel = _call(
+                Mix_PlayChannel,
+                -1,  # Auto-pick channel
+                chunk,
+                0,  # Do not repeat
+                _check_error=lambda rv: rv == -1
+            )
+            self._currently_playing[channel] = sound  # Keep reference of playing asset
```
astronouth7303, line 153: Can we move this to an `else` of the `try`? Just because I generally like to keep the `try` bodies as minimal as possible.

ppb/systems/sound.py
```diff
@@ -141,14 +142,18 @@ def on_play_sound(self, event, signal):
...
+        except SdlMixerError as e:
+            if not str(e).endswith("No free channels available"):
```
astronouth7303, line 155: Can we log or print a warning when this does happen?

viztests/sound_flooded_channels.py
```diff
@@ -0,0 +1,28 @@
+"""
```
astronouth7303, line 1: Thank you for including this.

ppb/systems/sound.py
pathunstrom: I literally had that thought after I submitted it.

PR #425, opened 2020-04-16
Fix Linux Installation

AstraLuma:
This removes the deviance of Linux from the other systems.

Blocked on https://github.com/a-hurst/pysdl2-dll/issues/2


AstraLuma:
Fixed in various other places.

PR #431, opened 2020-04-17
Use time.perf_counter for frame target calculation, more precise than…

ironfroggy:
… monotonic and explicitly used for comparisons.

### DIFF
--- a/ppb/utils.py
+++ b/ppb/utils.py
@@ -32,6 +32,14 @@ def camel_to_snake(txt):
 
 
 def get_time():
+    """
+    Returns the time via the default timer.
+    
+    Currently uses :func:`time.perf_counter` as the default timer.
+    
+    .. warning:: This is not a globally synchronized timer, it's just simply a system time. It is intended
+       to make sure all timers in ppb code use the same function.
+    """
     return perf_counter()
 
 
--- a/ppb/utils.py
+++ b/ppb/utils.py
@@ -1,8 +1,9 @@
 import logging
 import re
 import sys
+from time import perf_counter
 
-__all__ = 'LoggingMixin', 'camel_to_snake'
+__all__ = 'LoggingMixin', 'camel_to_snake', 'get_time'
 
 
 # Dictionary mapping file names -> module names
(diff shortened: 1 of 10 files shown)


astronouth7303:
ironfroggy on discord:

> [`perf_counter` is] defined as using the highest resolution clock the system has available. on *nix its actually the same underlying mechanism as `monotonic()`. Both are monotonic (no backwards). > On windows it uses a higher precision timer `QueryPerformanceCounter()` with a precision of 100 ns and a maximum value of 1.84467441E16 seconds (its a 64bit signed integer)
> i can back this all up across the python docs and source and MSDN docs on the internals if you're worried about the details. I need to file a ticket for this anyway and can include all the relevant links.

Given that there doesn't seem to be a downside to using `perf_counter()` everywhere, we should probably just use that?

ironfroggy:
@astronouth7303 I agree, and can expand this PR to change out the timer everywhere.

astronouth7303:
Yeah, please do that.

astronouth7303:
@ironfroggy ping

pathunstrom:
I'm not against it. We can give it a shot and if it feels more like clutter in the ppb namespace we can push it down to utils. I do like that ppb.utils.get_time() allows us to change what performance counter we're using without changing significant parts of the code base.

astronouth7303:
It's more the fact that it gives the illusion of a global synchronized time, which isn't really true

pathunstrom:
Oh, yeah, a docstring that explains where it's getting its time is probably a good thing.

ppb/utils.py
```diff
@@ -30,6 +31,10 @@ def camel_to_snake(txt):
     return _boundaries_finder_2.sub(r'\1_\2', s1).lower()
 
 
+def get_time():
+    return perf_counter()
```
pathunstrom, line 35:
```suggestion
def get_time():
    """
    Returns the time via the default timer.
    
    Currently uses :func:`time.perf_counter` as the default timer.
    
    .. warning:: This is not a globally synchronized timer, it's just simply a system time. It is intended
       to make sure all timers in ppb code use the same function.
    """
    return perf_counter()
```

Issue #373, opened 2019-09-30
Chained Assets


astronouth7303:
Extend the asset system to support chained assets--Virtual assets that depend on other assets and do additional background work.

Issue #419, opened 2020-04-12
Changing Sound asset volume before setup causes hang at startup


ironfroggy:
This works:

```
SOUND_CHIME = ppb.Sound("resources/sound/chimes.wav")
def setup(scene):
    SOUND_SWAP.volume = 0.5
    ...
ppb.run(setup=setup)
```

But this causes a permanent hang at startup:

```
SOUND_CHIME = ppb.Sound("resources/sound/chimes.wav")
SOUND_CHIME.volume = 0.5
def setup(scene):
    ...
ppb.run(setup=setup)
```

The only output I get is this message about the asset:

```
Waited on <Sound name='resources/sound/chimes.wav'> before the engine began
```

pathunstrom:
I know where that error comes from, but I don't know enough about the implementation to know if we can unhook the volume attribute from the asset load.

@astronouth7303 knows this bit of code.

astronouth7303:
OH yes. That was a dumb implementation.

Asset loads don't begin until the engine initializes, and the volume implementation depends on it.

This can be fixed without API changes.

Although upon review I wonder if this is a bad API because it'll change the volume for all usages and there's no way to create new copies.

pathunstrom:
Maybe make the volume be a parameter to the event and the sound system can duplicate at the place of use?

PR #432, opened 2020-04-18
Next Generation Assets

astronouth7303:
Or: If assetlib is so good, why isn't there assetlib 2?

This refactors/rewrites many aspects of assetlib to simplify implementations and better enable future asset features. (#373)

### DIFF
--- a/docs/reference/assets.rst
+++ b/docs/reference/assets.rst
@@ -34,6 +34,10 @@ system and the data logistics.
 
         Called in the background thread.
 
+    .. automethod:: load(timeout: float = None)
+
+    .. automethod:: is_loaded()
+
 
 Subclassing
 ~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


astronouth7303:
Ok, did a quick docs fixup (although the page needs a real pass).

I think this is ready for review.

astronouth7303:
Umm... it really shouldn't be that noisy.

That warning that covers cases like #419, where someone tries to call `.load()` before the engine starts, which should result in deadlock.

Given that the lifespan of the executor should match the lifespan of the system, I'm not certain how this happened.

(I could, of course, refine the conditions in which the warning is given, but i'm still extremely puzzled as to why it happened.)

pathunstrom:
My guess is that I define most of my assets at either the module level or the class attribute level, so when I import the first scene, it chains in basically everything else, and that happens before the engine starts?

astronouth7303:
The warnings came from `FreeingMixin.__del__()` conflicting `BackgroundMixin.load()`'s warning. There isn't a real fix without a more specific combination.

Fixed by shushing if the asset is already loaded.

PR #436, opened 2020-04-18
Assets 2: Chained assets

### LINKED ISSUE #373: Chained Assets

astronouth7303:
Depends on #432 

Implements chained assets: Assets that depend on other assets (such as how Text depends on Fonts)

Closes #373

### DIFF
--- a/tests/test_assets.py
+++ b/tests/test_assets.py
@@ -6,7 +6,10 @@ import pytest
 from ppb import GameEngine, BaseScene
 import ppb.events
 import ppb.assetlib
-from ppb.assetlib import DelayedThreadExecutor, Asset, AssetLoadingSystem
+from ppb.assetlib import (
+    DelayedThreadExecutor, Asset, AssetLoadingSystem, BackgroundMixin,
+    ChainingMixin, AbstractAsset,
+)
 from ppb.testutils import Failer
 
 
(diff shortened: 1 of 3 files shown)


tests/test_assets.py
```diff
@@ -181,3 +184,63 @@ def test_timeout(clean_assets):
 
     with pytest.raises(concurrent.futures.TimeoutError):
         a.load(timeout=0.1)
+
+
+def test_chained(clean_assets):
+    class Const(BackgroundMixin, AbstractAsset):
```
pathunstrom, line 190: This is kind of magical, can we get a comment that this is essentially a threaded string.join so folks at least get the intent?

tests/test_assets.py
```diff
@@ -181,3 +184,63 @@ def test_timeout(clean_assets):
...
+        )
+    )
+    engine = GameEngine(
+        AssetTestScene, basic_systems=[AssetLoadingSystem, Failer],
+        fail=lambda e: False, message=None, run_time=1,
+    )
+    with engine:
+        engine.start()
+
+        assert a.load(timeout=5)
```
pathunstrom, line 246: I'm not sure what this is intended to test? Just that all of these workers are loaded in less than 5 seconds?

astronouth7303: It's so that if the chaining deadlocks, the test fails instead of pytest just sitting.

(Yes, this happened quite a bit in development.)

Issue #400, opened 2020-03-25
Refresh BaseScene arguments


astronouth7303:
Looking at `BaseScene`:

```
    def __init__(self, *,
                 set_up: Callable = None, pixel_ratio: Number = 64,
                 **kwargs):
```

* `set_up` should be handled by `ppb.run()` or `Engine`, not the scene
* `pixel_ratio` seems like it should be handed to the camera by the Renderer, not passed through the scene.

pathunstrom:
I think this should be split into two tickets (setup's change futzes with the engine, the pixel ratio one futzes with the renderer, so very different places in the code base.)

PR #439, opened 2020-04-19
Camera Redesign

### LINKED ISSUE #185: Change the various camera relationships.
Currently, frame width and height are derived from the viewport size and the pixel ratio. This makes calculating what counts as _in_ much harder for the end user.

The viewport should have a screen origin, a percentage of the screen it occupies in each axis, and knowledge of the resolution. It then derives it's width and height from these factors.

The frame should have a width and a height. It should have a method available for forcing the frame aspect ratio to match the viewport aspect ratio, which can be called when any of the root values is changed: the sceen percentages, the resolution, or the frame width (Or frame height in the case that a user sets the zoom using it.)

### LINKED ISSUE #310: Incomplete camera information at scene init
When a scene is initialized, its `Camera` has not received information from the renderer yet, making some forms of pinning.

This is true whether it's the initial scene, a stacked scene, or replacing the initial scene.

pathunstrom:
This is the long awaited fix to all the pain points in the Camera. So far, it's just an __init__.

This represents a breaking change in the API.

This touches on a number of issues, some solved wholly by this change, some only fixes part of the problem.

Resolves #185: pixel_ratio is now derived from the number of game units you want visible and the screen resolution. You can "zoom" by increasing or decreasing the width or height of the camera.
Resolves #310: Camera isn't added until SceneStarted event, but is fully initialized as soon as it's available.
Regarding #358: Did not deal with this directly. That test is currently nullified until we can go over it and determine what is going on.
Regarding #400:  Removes the pixel_ratio argument from BaseScene.


docs/reference/camera.rst
```diff
@@ -0,0 +1,18 @@
...
+.. autoclass:: Camera
+   :members:
+
+
+.. warning::
+   Setting the game unit dimensions of a camera (whether via
+   :attr:`Camera.width`, :attr:`Camera.height`, or the
+   ``target_game_unit_width`` of the :class:`Camera` constructor) will
+   affect both :attr:`Camera.width` and :attr:`Camera.height`. Their ratio
+   is determined by the defined window.
```
astronouth7303, line 18: I think we might need more prose to discuss camera initialization and when scenes can expect to have the camera.

ppb/systems/renderer.py
```diff
@@ -111,12 +113,14 @@ def __init__(
         resolution=DEFAULT_RESOLUTION,
         window_title: str = "PursuedPyBear",
         target_frame_rate: int = 30,
+        target_camera_width=25,
```
astronouth7303, line 116: This is the default scene width? This is significantly larger than before, right?

ppb/systems/renderer.py
```diff
@@ -152,15 +156,23 @@ def __exit__(self, *exc):
...
-        self.pixel_ratio = camera.pixel_ratio
+    def on_scene_started(self, scene_started, signal):
+        scene = scene_started.scene
+        camera_class = getattr(scene_started.scene, "camera_class", Camera)
+        # For future: This is basically the pattern we'd use to define
+        # multiple cameras. We'd just need to have the scene tell us the
+        # regions they should render to.
+        camera = camera_class(self, self.target_camera_width, self.resolution)
+        scene.main_camera = camera
+        self.scene_cameras[scene] = [camera]
```
astronouth7303, line 171: Why the extra ref?

ppb/systems/renderer.py
```diff
@@ -234,9 +246,10 @@ def compute_rectangles(self, texture, game_object, camera):
 
         return src_rect, dest_rect, ctypes.c_double(-game_object.rotation)
 
-    def target_resolution(self, width, height, game_unit_size):
+    @staticmethod
+    def target_resolution(width, height, game_object_size, pixel_ratio):
         values = [width, height]
         short_side_index = width > height
-        target = self.pixel_ratio * game_unit_size
+        target = pixel_ratio * game_object_size
```
astronouth7303, line 253: Not actually a fan of this rename. The old name was more correct and we're not moving away from game units as a concept.

tests/test_camera.py
```diff
@@ -1,130 +1,165 @@
...
+    [Vector(0, 0), Vector(320, 380), Vector(-1, -1)],
+    [Vector(0, 0), Vector(720, 540), Vector(4, -3)],
+    [Vector(0, 0), Vector(-80, -20), Vector(-6, 4)]
+])
+def test_camera_translate_point_to_game_space(camera, position, point, expected):
+    camera.position = position
+    assert camera.translate_point_to_game_space(point) == expected
+
+
+# @pytest.mark.skip(reason="Test for old camera. Will want to restore this functionality in new camera.")
```
astronouth7303, line 107: Any reason these tests weren't ported?

ppb/systems/renderer.py
pathunstrom: Flipped a coin between a 32 pixel default sprite and a 64. I'm willing to switch it.

ppb/systems/renderer.py
pathunstrom: Primarily to have handles to adjust resolution when we finally start supporting resizeable windows.

ppb/systems/renderer.py
pathunstrom: Oh, yeah, sorry, I inlined this because pixel_ratio isn't directly a renderer concern anymore, then realized we had a test around this. Can revert.

tests/test_camera.py
pathunstrom: Well, this top one is because I didn't rewrite `sprite_in_viewport` because it's not used by anything in the library. I'll probably do it sometime in the next couple of months and then restore/rewrite the relevant test.

The one below it is because it's doing a lot, would require translation to the new camera tests, and I'm not entirely sure what all of the steps are there to do. As such, I'd rather rewrite it as its own thing and not tied to the giant pile of other changes going on. (And it's commented out because the old camera got killed, I'd have just left it marked skip otherwise.)

ppb/systems/renderer.py
astronouth7303: Yes please.

tests/test_camera.py
astronouth7303: I think that's a useful utility, although the name isn't great.

The one below that is supposed to be a simple round-trip test: If we convert a point from game units to pixels and back, do we get the same point?

pathunstrom: It is a useful utility! I have no intention of leaving it cut forever. I just wanted to get the new camera _working_ then I'll restore some of the niceties later.

The good news RE: round tripping, the tests for translate_point_to_screen and translate_point_to_game_space use the same values for their tests, just reversing input and output.

astronouth7303: Well, doing a roundtrip test means it's generic and we can throw hypothesis at it.

astronouth7303:
We can drop `sprite_in_viewport()` for now, but i'll hold you to rewriting it.

ppb/systems/renderer.py
pathunstrom: (But yes, this is 25 x ~19, previously, we were 10 x ~7.5)

PR #451, opened 2020-04-26
Adds documentation of ppb.__init__.

pathunstrom:
Adds a detailed package docstring with the basic pieces of the library called out.
Adds documentation to `run` and `make_engine`.
Removes `starting.rst` in favor of including those examples in the `run()` documentation.

Screen shot:

init-docs


docs/reference/index.rst
```diff
@@ -1,15 +1,22 @@
...
 ================
 
 For as simple as the tutorials make ``ppb`` look there's a lot of power under
 the hood. This section will cover the raw what of the ppb API. To find out why
-decisions are made, see the discussion section.
+decisions are made, see the :doc:`/discussion/index` section.
+
+.. automodule:: ppb
+
+.. autofunction:: run
```
astronouth7303, line 12: Did you build the docs locally? I think these two should be indented.

pathunstrom: Yup, looked fine to me? Automodule just prints out the docstring as text, it doesn't give a heading the same way `.. autoclass` or `.. autofunction` do.

Give me a second and I'll post a screenshot.

Issue #382, opened 2020-02-28
Default Sprite Behaviors: TargetSprite


pathunstrom:
In the interest of improving the on-boarding ramp for new users and learners, and by extension making our live demonstrations a little more impressive, we'd like to include some Sprites with basic control features.

For this ticket we are defining a `TargetSprite`. A `TargetSprite` will function very similarly to a `FollowerSprite` #381, but target a Vector instead of an object. The primary difference between these two is that a TargetSprite has an immutable Vector as its target, so does not shift it's direction of movement unless it's updated via an event handler or modified by a system.

We'd like to provide a collection of movement strategies: various kinds of linear interpolation and a basic physics model with a velocity, acceleration and max velocity.

Due to its nature of requiring additional behaviors to change its target, `TargetSprite` is intended to be subclassed.

PR #456, opened 2020-05-06
(WIP) TargetSprite

serin-delaunay:
Class `TargetSprite` with parameters `target` and `speed`. Moves towards target at the given speed, and stops there. Will resolve https://github.com/ppb/pursuedpybear/issues/382.

### DIFF
--- a/ppb/features/default_sprites.py
+++ b/ppb/features/default_sprites.py
@@ -7,5 +7,10 @@ class TargetSprite(ppb.BaseSprite):
     speed = 1.0
 
     def on_update(self, update_event, signal):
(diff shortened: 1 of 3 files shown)


ppb/features/default_sprites.py
```diff
@@ -0,0 +1,43 @@
+import ppb
+import math
+
+
+class TargetSprite(ppb.sprites.BaseSprite):
```
pathunstrom, line 5:
```suggestion
class TargetSprite(ppb.Sprite):
```

Just to make sure all of the expected features are available.

PR #458, opened 2020-05-07
Saften TargetSprite

serin-delaunay:
* Crash on negative exponential_speed to prevent imaginary values.
* Document effects of parameter value ranges.
* Prevent zero division when max speed is negative and sprite is at target.

Tests still only cover the standard usage, not error conditions or negative speeds.

### DIFF
--- a/ppb/features/default_sprites.py
+++ b/ppb/features/default_sprites.py
@@ -26,7 +26,7 @@ class TargetSprite(ppb.Sprite):
         if self.max_speed < self.min_speed:
             raise ValueError("TargetSprite maximum speed cannot be less than minimum speed.")
         if self.exponential_speed > 1.0:
-            raise ValueError("TargetSprite exponential speed cannot be greater than 1.")
+            raise ValueError(f"{type(self).__name__} exponential speed cannot be greater than 1.")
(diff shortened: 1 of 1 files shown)


ppb/features/default_sprites.py
```diff
@@ -21,8 +25,13 @@ class TargetSprite(ppb.Sprite):
     def on_update(self, update_event, signal):
         if self.max_speed < self.min_speed:
             raise ValueError("TargetSprite maximum speed cannot be less than minimum speed.")
+        if self.exponential_speed > 1.0:
+            raise ValueError("TargetSprite exponential speed cannot be greater than 1.")
```
pathunstrom, line 29:
```suggestion
            raise ValueError(f"{type(self).__name__} exponential speed cannot be greater than 1.")
```

This will let subclassed versions retain their class name in the error message.

serin-delaunay: Did the same for the max/min speed error.

pathunstrom:
Just want to ping this. I'm happy with what you've done here, just need it to be not a draft to get it in to the code base.

pathunstrom:
bors retry

PR #463, opened 2020-05-09
Adds ordinal direction unit vectors.

pathunstrom:
Primary use case is to reduce the cognitive load in early lessons:

```python
class OurSprite(Sprite):
    def on_update():
       self.facing = Left
```


pathunstrom:
I'll change the casing, figure ppb.directions.Left is probably a good place to start and be open to including it in the core name space if it's requested by teachers?

astronouth7303:
Good enough explanation.

PR #466, opened 2020-05-11
Rectangular sprite rendering

### LINKED ISSUE #356: Renderer should ignore sprites without __image__
The current renderer will choke if a sprite doesn't have an `__image__` method. During iteration of the scene, the renderer should ignore sprites without one.

astronouth7303:
Handles rendering for #376 

Closes #356


ppb/systems/renderer.py
```diff
@@ -202,8 +202,20 @@ def render_background(self, scene):
         )
         sdl_call(SDL_RenderClear, self.renderer, _check_error=lambda rv: rv < 0)
 
+    def is_visible(self, game_object):
+        if hasattr(game_object, 'width') and game_object.width <= 0:
```
pathunstrom, line 206: Would it be valuable to also exclude Sprites without `__image__` or `__image__() == None` here?

astronouth7303: Ehhhh..... Maybe if we expected it to be reused? I'd honestly would rather just rename it, since it's a single-use helper and the one call site immediately calls `__image__()` right after.

pathunstrom: But doesn't it also check DoNotRender there? We could bail earlier.

astronouth7303: Oh, right, the only difference would be if it didn't have `__image__`.

pathunstrom:
As a note, I think with 0fada4e this resolves #356.

PR #474, opened 2020-05-17
Apply alpha blend modes and coloration to textures based on sprite data.

ironfroggy:
I know there was debate before about where `opacity` and `color` data should live, but i wanted to open the PR with the customizations I've been using to start the discussion again with something concrete.

### DIFF
--- a/ppb/systems/renderer.py
+++ b/ppb/systems/renderer.py
@@ -2,6 +2,7 @@ import ctypes
 import io
 import logging
 import random
+from typing import Tuple
 
 import sdl2
 import sdl2.ext
@@ -25,6 +26,11 @@ from sdl2 import (
     SDL_QueryTexture,  # https://wiki.libsdl.org/SDL_QueryTexture
     SDL_RenderCopyEx,  # https://wiki.libsdl.org/SDL_RenderCopyEx
     SDL_CreateRGBSurface,  # https://wiki.libsdl.org/SDL_CreateRGBSurface
+    SDL_BLENDMODE_ADD,
+    SDL_BLENDMODE_BLEND,
+    SDL_SetTextureAlphaMod,
+    SDL_SetTextureBlendMode,
+    SDL_SetTextureColorMod,
 )
 
 from sdl2.sdlimage import (
@@ -107,6 +113,9 @@ class SmartPointer:
 
 class Renderer(SdlSubSystem):
     _sdl_subsystems = SDL_INIT_VIDEO
+    last_opacity: int = 255
+    last_opacity_mode: str = 'blend'
+    last_color: Tuple[int, int, int] = (255, 255, 255)
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


ppb/systems/renderer.py
```diff
@@ -107,6 +113,9 @@ def __del__(self):
 
 class Renderer(SdlSubSystem):
     _sdl_subsystems = SDL_INIT_VIDEO
+    last_opacity: int = 255
+    last_opacity_mode: str = 'blend'
+    last_color: Tuple[int, int, int] = (255, 255, 255)
```
pathunstrom, line 118: Are these doing anything? I don't see them referenced elsewhere?

ppb/systems/renderer.py
```diff
@@ -233,14 +242,41 @@ def prepare_resource(self, game_object):
...
+
+        opacity = getattr(game_object, 'opacity', 255)
+        opacity_mode = getattr(game_object, 'opacity_mode', 'blend')
+        color = getattr(game_object, 'color', (255, 255, 255))
+
+        sdl_call(
+            SDL_SetTextureAlphaMod, texture.inner, opacity,
+            _check_error=lambda rv: rv < 0
+        )
+        if opacity_mode == 'add':
```
pathunstrom, line 261: Direct implementation is less of a big deal, but I'd definitely avoid magic strings if we can. Either make these class attributes, exported constants, or Flags. (Or enums, but want to check with some educators before we go too far down that route.)

ppb/systems/renderer.py
ironfroggy: this was part of an attempt I made to only call the SDL `Set*` functions when these changed, but I didn't get it working right and dropped it. I will clean this up.

pathunstrom: Was just checking!

ironfroggy:
Here's an example of where I use this in _Seed Magic_ to do the particle effects:

```python
def spawn(cls, pos, color, heading=None, tsize=2.5):
        s = cls.sparkles[cls.index]
        cls.index = (cls.index + 1) % cls.size
        if color == COLOR_BLACK:
            s.opacity_mode = ppb.flag.BlendModeNone
            s.opacity = 255
        else:
            s.opacity_mode = ppb.flag.BlendModeAdd
            s.opacity = 128
        s.color = color
        s.position = pos
        s.rotation = randint(0, 260)
        s.size = 1.5
        s.layer = 100
        cls.t.tween(s, 'opacity', 0, 0.5, easing='linear')
        cls.t.tween(s, 'size', tsize, 0.5, easing='linear')
        delay(0.5, lambda: setattr(s, 'size', 0))
        if heading:
            cls.t.tween(s, 'position', heading, 0.5, easing='linear')
```

pathunstrom:
One more question on this, is `Sprite.color` essentially for tinting the underlying images?

ironfroggy:
@pathunstrom Yes, right. Actually, with this, the generated sprites could be recolored on-demand, which might be useful.

ppb/systems/renderer.py
```diff
@@ -233,14 +248,35 @@ def prepare_resource(self, game_object):
...
                 SDL_CreateTextureFromSurface, self.renderer, surface,
                 _check_error=lambda rv: not rv
             ), SDL_DestroyTexture)
             self._texture_cache[surface] = texture
-            return texture
+
+        opacity = getattr(game_object, 'opacity', 255)
+        opacity_mode = getattr(game_object, 'opacity_mode', flags.BlendModeBlend)
+        opacity_mode = OPACITY_MODES[opacity_mode]
+        color = getattr(game_object, 'color', (255, 255, 255))
```
pathunstrom, line 262:
```suggestion
        color = getattr(game_object, 'tint', (255, 255, 255))
```

Regarding suggestion on discord.

pathunstrom:
bors retry

PR #475, opened 2020-05-17
Support Sprite.rect property, a 4-tuple controlling whthe src-rect to…

ironfroggy:
… render from the texture.

### DIFF
--- a/ppb/systems/renderer.py
+++ b/ppb/systems/renderer.py
@@ -255,13 +255,19 @@ class Renderer(SdlSubSystem):
 
         src_rect = SDL_Rect(x=0, y=0, w=img_w, h=img_h)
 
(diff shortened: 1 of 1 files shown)


ironfroggy:
If this or a version of this is merged, I'd also like to contribute a `ppb.features.spritesheet` to make use of it with traditional spritesheets easier. Automatic slicing, that sort of thing.

pathunstrom:
I like the idea of this, but I think we need to figure out some of the stuff around this API, so going to say let's have the conversation, and we'll make figuring this out a priority for the 0.10 work. (Not going to close it though, because this might be pretty close to the renderer implementation no matter how the rest of the discussion goes.)

pathunstrom:
Going to connect this to #501 and the discussion around the future of assets.

PR #471, opened 2020-05-15
Docs sprites

pathunstrom:

PR #476, opened 2020-05-18
Have CI sphinx treat warnings as errors

astronouth7303:
We're accumulating sphinx warnings, which usually means that sphinx is doing something wacky in the generated output.

Have the CI docs check treat these as errors so they don't get into the code.

Depends on #471.


astronouth7303:
@pathunstrom this seems ready

pathunstrom:
bors retry

pathunstrom:
bors cancel

PR #483, opened 2020-05-24
[WIP] Various render and event loop optimizations to share

ironfroggy:
Not intended as a mergeable pull request (yet?) but a starting point for conversations on event and render loop design.

All these changes got my ~1000 sprite prototype from ~40 fps to ~100 fps

### DIFF
--- a/ppb/assetlib.py
+++ b/ppb/assetlib.py
@@ -212,9 +212,13 @@ class BackgroundMixin:
         Will block until the data is loaded.
         """
         # NOTE: This is called by FreeingMixin.__del__()
-        if not self.is_loaded() and not _executor.running():
-            logger.warning(f"Waited on {self!r} outside of the engine")
-        return self._future.result(timeout)
+        try:
+            return self._cached_result
+        except AttributeError:
+            if not self.is_loaded() and not _executor.running():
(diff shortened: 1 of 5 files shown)


ppb/assetlib.py
```diff
@@ -212,9 +212,13 @@ def load(self, timeout: float = None):
         Will block until the data is loaded.
         """
         # NOTE: This is called by FreeingMixin.__del__()
-        if not self.is_loaded() and not _executor.running():
-            logger.warning(f"Waited on {self!r} outside of the engine")
-        return self._future.result(timeout)
+        try:
+            return self._cached_result
```
pathunstrom, line 216: We should benchmark this independently. We used to do something like this in Vector and found just doing the work was cheaper than the cost of the look-ups.

ironfroggy: I definitely would want to benchmark this, but I will at least point out that in any case where we want to cache something doing it with a try/except block is cheaper than a lookup followed by a calculation because it incurs 0 costs on the hit branch. The try-except variant does not _have_ a cost of look-ups, necessarily.

pathunstrom:
Noting I'm looking to start poking the various optimizations here and we can start breaking this into separate changes.

PR #489, opened 2020-06-06
Fixes Text Rendering using 1 Font object

### LINKED ISSUE #488: Reusing a Font for multiple Text assets can cause crashes.
```python
import ppb

font = ppb.Font("my_font", size=72)
my_first_text = ppb.Text("My first text", font=font, color=(255, 255, 255))
my_second_text = ppb.Text("My second text", font=font, color=(255, 255, 255))

def setup(scene):
    scene.add(ppb.Sprite(image=my_first_text))
    scene.add(ppb.Sprite(image=my_second_text))

ppb.run(setup)
```

This sample will fail, somewhat randomly, with multiple possible errors depending the order the various Text assets background threads get completed.

Can be fixed by only calling the SDL calls with the lock.

(Fix incoming.)

pathunstrom:
Needs to use the mutex around FT_Libary.

Closes #488 

Co-authored-by: Jamie Bliss <jamie@ivyleav.es>

### DIFF
--- a/ppb/systems/text.py
+++ b/ppb/systems/text.py
@@ -29,6 +29,9 @@ from ppb.systems._sdl_utils import ttf_call
 # used around FT_New_Face and FT_Done_Face.
 #
 # I assume this translates to TTF_OpenFont* and TTF_CloseFont
+# SDL_ttf manages a single global FT_Library, so we need to use the lock
+# for threaded calls into it, like in Asset._background.
+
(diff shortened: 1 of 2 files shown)


astronouth7303:
We should look into making that test automated.

pathunstrom:
Long term, I agree. Tonight, do you think that'll be trivial?

PR #500, opened 2020-07-11
Implement the Game Object Model

AstraLuma:
Implements a standard interface for "game objects" (things that participate in the tree and receive events)

Part of #492

PR #501, opened 2020-07-16
Rendering context: A place to hold rendering related values

pathunstrom:
The basic idea is simple: Instead of sprites growing more and more data, we store them on a context object.

As we continue to add rendering options, we've been adding more and more symbols to the RenderableMixin. This creates an issue as we are taking up simple words end users might want to use for other purposes. This is a proposal for a way to manage this via a context object: A dataclass that holds all of the relevant details the Renderer cares about.

This PR tries to maintain backwards compatibility with the previous model, but if we go this route we can discuss how best to sunset the old patterns going forward.

## Things to continue supporting

```python
Sprite.image = ImageAsset
```

This should continue to work. It current does at definition time, but after the rendering_info exists, this stops working. Want to think on the best way to implement that before we do.

## Nice to haves

I would like to make simple access available still, but doing that could explode the complexity of this API and create a problem where we're adding a bunch of tokens to sprites, plus our new one. Maybe we could ship a feature that takes the things in RenderInfo and adds pass through properties as a mixin? Supports a best of both worlds, without actually requiring all of those attributes be used the way we want them to be, and lets end users opt in to adding to their name spaces.

## Problems

This PR represents a new type of ppb object (`contexts`) and I'd like us to examine if this is the best way. They're just dataclasses, which means minimal maintenance, but added complexity and moving us closer to a true ECS. (Which. . . let's be real #500 is already moving us closer to that world anyway.)

Open to comments will update this document to make sure the discussion is encapsulated well.

### DIFF
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -33,6 +33,7 @@ import logging
 import warnings
 from typing import Callable
 
+from ppb import contexts
 from ppb import directions
 from ppb import events
 from ppb_vector import Vector
@@ -53,6 +54,7 @@ __all__ = (
     # Shortcuts
     'Vector', 'BaseScene', 'Circle', 'Image', 'Sprite', 'RectangleSprite',
     'Square', 'Sound', 'Triangle', 'events', 'Font', 'Text', 'directions',
+    'contexts',
     # Local stuff
     'run', 'make_engine',
 )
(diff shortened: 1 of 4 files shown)


AstraLuma:
This reads similar to some of my ideas about pixmap/pixbuf objects that take ownership of surfaces, textures, and rendering options?

pathunstrom:
Possibly. This might even just be a halfway point to the pixbuff world, and I'd be okay with that.

The most complicated thing I'm seeing in the API design is that we still want the simple cases to work, and I'm not convinced the old `__image__` protocol is going to be the right way to do that.

PR #506, opened 2020-08-13
Fix docs

AstraLuma:
Fix docs


pathunstrom:
bors r+

I realize we should probably verify out type annotations as part of CI, too.

Issue #492, opened 2020-06-19
Universal Tree Model/Game Object Model


AstraLuma:
This is something we've been discussing a long time, but it hasn't had it's own discussion yet.

The idea is to unify the "game object" interface that defines the object tree (engine->systems,scenes->sprites), as well as enable some additional features.

In addition to making tree traversal simpler, some of the additional features I'd like to include:
* Allow for off-interface Actors (eg, non-renderables under scenes, anything under sprites or systems)
* Simpler/more consistent handling of subsets of children (eg, sprites)

No, I don't know if I want to call it the Unified Tree Model, the Game Object Model, or something else. Suggestions welcome.

astronouth7303:
I think I want to define a `GameObject` mixin that defines some standard children handling and such, as well as some child subset utility classes. And handle our semi-standard property-based inits.

Then everything that can exist in the tree will inherit from it.

Or, and to clarify: The biggest use of the tree is events, but it also influences things like scene scope, cleanup, etc. (Like, if you parent a tween to a sprite, when you remove the sprite, the tweens go to.)

Things I'm not going to do this time: Sub sprites, sprite groups. Subsprites still need some time to percolate, and I think I would like to think about sprite groups along with some of the discussions around non-event/non-actor sprites.

pathunstrom:
I think Game Object Model is more reasonable long term. We get the UTM idea, but when I imagine the documentation about this, I think GOM is a better title.

pathunstrom:
I think we can consider this one complete?

AstraLuma:
Probably, yeah

PR #507, opened 2020-08-13
GOM: Update the event loop

AstraLuma:
Update the event loop to use the GOM

Part of #492


pathunstrom:
So if I'm reading this correctly: This is a non-breaking change? Since end user code won't need to change at all to support this.

AstraLuma:
Yeah, this only generalizes the tree and enables new tree shapes.

pathunstrom:
Looks good.

bors r+

PR #516, opened 2020-09-01
Allow engine to be instantiated with an instantiated scene -- Closes #335

### LINKED ISSUE #335: Allow Engine.starting_scene to be an instantiated scene
Currently, you cannot pass an instantiated scene to the `GameEngine` (in any of the standard starting patterns). You can only pass a class and kwargs.

This should be allowed.

mfonism:
This PR makes it possible to pass an instantiated scene to a game engine as per #335

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -14,6 +14,7 @@ from ppb import events
 from ppb.assetlib import AssetLoadingSystem
 from ppb.gomlib import walk
 from ppb.errors import BadEventHandlerException
+from ppb.scenes import BaseScene
 from ppb.systems import EventPoller
 from ppb.systems import Renderer
 from ppb.systems import SoundController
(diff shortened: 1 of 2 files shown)


mfonism:
@AstraLuma, @pathunstrom is this the right way to go?

I implemented duck-typing on the base scene class because doing an instance check in the engine class caused this test to fail as `mock_scene_class` is seen as an instance of `ppb.scenes.BaseScene` (because of the `spec` arg in the mock, I think).

https://github.com/ppb/pursuedpybear/blob/68f4ee9dde242c61b658786b580a4a5f66f4794a/tests/test_engine.py#L16-L22

mfonism:
If this is the right way to go, I can go ahead to update the affected doc strings and type annotations.

AstraLuma:
No, it would be better to just do an instance check.

~Not sure why you're using a mock for testing? You're not using any of the features a mock provides, and scenes are memory-only data structures.~ nvm, the mock usage was ours. your test looks good.

pathunstrom:
I think I agree with @AstraLuma that the right way to handle this is an instance check against `type`. This solution was clever, but has a possibility of creating confusion elsewhere.

I'd keep your existing test, it's great. What I'd like to see is the old test rewritten (if you're up for it) to just make sure that after engine.start we have a functioning scene, and parameterize the test to take BaseScene, BaseScene(), maybe an additional mock. Replace the assert with checking the length of engine.children equals 1.

Then you should be able to do the instance check without problem and we keep the existing test.

mfonism:
@pathunstrom are there any more improvements I can make to the code?

Can I proceed to work on the affected doc strings and type annotations?

pathunstrom:
This looks good, definitely move on to docs and type annotations!

pathunstrom (approved):
All right, this all looks good.

bors r+

PR #519, opened 2020-09-07
GameObject: Implement on Engine

AstraLuma:
This does two things:

* Moves the management of children out of `Engine`
* Adds `GameObject` to `Engine`.

Part of #492

### DIFF
--- a/ppb/engine.py
+++ b/ppb/engine.py
@@ -6,13 +6,20 @@ from itertools import chain
 from typing import Any
 from typing import Callable
 from typing import DefaultDict
+from typing import Hashable
+from typing import Iterable
+from typing import Iterator
 from typing import List
 from typing import Type
 from typing import Union
 
+import ppb
+import ppb.systemslib
 from ppb import events
 from ppb.assetlib import AssetLoadingSystem
+from ppb.gomlib import Children, GameObject
 from ppb.gomlib import walk
+from ppb.errors import BadChildException
 from ppb.errors import BadEventHandlerException
 from ppb.systems import EventPoller
 from ppb.systems import Renderer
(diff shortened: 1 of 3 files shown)


AstraLuma:
~Currenly viztests are failing for me in Wacky Ways, so I haven't tested on those.~ yup it works.

ppb/engine.py
```diff
@@ -39,7 +46,170 @@ def _get_handler_name(txt):
...
+
+        Note that Scenes and Systems have special restrictions.
+
+        Examples: ::
+
+            children.add(MyObject())
+
+            children.add(MyObject(), tags=("red", "blue")
+        """
+        # Ugh, this is a copy of the implementation in Children.
```
pathunstrom, line 116: Not for this PR but we could consider a "validate_child" helper function in `Children` that can be called in implementations of `add` and overridden for things like the check on line 122.

Need to consider the implementation considering we don't want to rewrite 131-134 either.

ppb/engine.py
```diff
@@ -70,25 +240,23 @@ def __init__(self, first_scene: Type, *,
            consequences. Consider passing via systems parameter instead.
         """
 
-        super(GameEngine, self).__init__()
+        super().__init__()  # FIXME: This is breaking the GameObject protocol
```
pathunstrom, line 244: O.o How is this breaking the protocol?

AstraLuma: Part of `GameObject` is that `__init__()` takes initial properties as kwargs.

```python
ge = GameEngine(spam='eggs')
ge.spam  # AttributeError
```

pathunstrom: Aha. I should play with run because we might be able to make that happen.

AstraLuma: It shouldn't be at the `run()` level? like, that's a convenience to make it easier to stand up an Engine.

pathunstrom: I mean, we convert the engine to only kwargs, run keeps its current API, we use run as the translation layer between.

pathunstrom (approved):
bors r+

PR #520, opened 2020-09-08
Remake Run

pathunstrom:
Per conversations between myself and @AstraLuma, we'd like to not pass all the kwargs down into the engine and the scenes, so we're pulling scene instantiation to the run function.

Hit a bug in the soundonly viztest, so won't get this done in time for feature freeze. As such, WIP, targeting 0.11.

Also, would love @moshez opinion as this will most drastically affect txppb outside of the library itself.

### DIFF
--- a/ppb/__init__.py
+++ b/ppb/__init__.py
@@ -32,6 +32,7 @@ Exports:
 import logging
 import warnings
 from typing import Callable
+from typing import Union
 
 from ppb import directions
 from ppb import events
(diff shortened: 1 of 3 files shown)


ppb/__init__.py
```diff
@@ -58,20 +59,18 @@
...
         "window_title": title,
         **engine_opts
     }
     return kwargs
 
 
-def run(setup: Callable[[BaseScene], None] = None, *, log_level=logging.WARNING,
-        starting_scene=BaseScene, title="PursuedPyBear", **engine_opts):
+def run(setup: Union[Callable[[BaseScene], None], BaseScene] = lambda scene: None, *, log_level=logging.WARNING,
+        starting_scene=None, scene_class=None, title="PursuedPyBear", **engine_opts):
```
AstraLuma, line 73: This seems like a silly deprecation?

I think the rename of `starting_scene` -> `scene_class` reduces clarity (slightly) and doesn't represent any meaningful improvement?

pathunstrom: Fair. I think my issue was the name shadowing, but reasonable.

PR #527, opened 2020-09-23
Re-ordered call to avoid OGG problem?

bildzeitung:
Possibly this: https://discourse.libsdl.org/t/bug-sdl2-mixer-init-error-format-support-not-available/23705/5


pathunstrom:
Noting the test failure isn't you, and this is currently blocked on the release which should be done tonight.

PR #538, opened 2020-10-16
Targeted events

### LINKED ISSUE #249: Targeted Events

AstraLuma:
Closes #249


pathunstrom (approved):
Love it.

bors r+

PR #541, opened 2020-10-16
gomlib: use deque instead of list for walk()

AstraLuma:
Exactly what it says on the tin.

This should be a performance improvement.

### DIFF
--- a/ppb/gomlib.py
+++ b/ppb/gomlib.py
@@ -208,7 +208,7 @@ def walk(root):
 
     Is non-recursive.
     """
-    q = deque(root)
+    q = deque([root])
     while q:
         cur = q.popleft()
         yield cur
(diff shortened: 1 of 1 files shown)


ppb/gomlib.py
```diff
@@ -208,9 +208,9 @@ def walk(root):
 
     Is non-recursive.
     """
-    q = [root]
+    q = deque(root)
     while q:
-        cur = q.pop(0)
+        cur = q.pop()
```
pathunstrom: Pretty sure this should be `pop_left`?

pathunstrom (approved):
bors r+

AstraLuma:
bors retry

ppb/gomlib.py
```diff
@@ -208,9 +208,9 @@ def walk(root):
 
     Is non-recursive.
     """
-    q = [root]
+    q = deque(root)
```
pathunstrom:
```suggestion
    q = deque([root])
```

Issue #517, opened 2020-09-03
Fix the use of deprecated attributes in tests


mfonism:
Running the tests gives the following warnings summary:

```
tests/test_gom.py::test_remove_methods[player0-container0]
tests/test_gom.py::test_remove_methods[player1-container1]
  ...\pursuedpybear\ppb\gomlib.py:187: DeprecationWarning: .kinds is deprecated, use .children.kinds()
    warnings.warn(".kinds is deprecated, use .children.kinds()", DeprecationWarning)

tests/test_gom.py::test_remove_methods[player0-container0]
tests/test_gom.py::test_remove_methods[player1-container1]
  ...\pursuedpybear\ppb\gomlib.py:199: DeprecationWarning: .tags is deprecated, use .children.tags()
    warnings.warn(".tags is deprecated, use .children.tags()", DeprecationWarning)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

We should fix this to avoid bad behaviour when the deprecations eventually kick in.

pathunstrom:
Agree. This is mostly a leftover as the GOM is implemented (#500, #507, #519).

PR #557, opened 2020-10-31
Fix the use of deprecated attributes in tests

mfonism:
This PR fixes the use of deprecated attributes in tests as per #517


AstraLuma:
bors try

PR #576, opened 2021-02-11
Add rectangles as a primitive asset.

### LINKED ISSUE #566: Create a Rectangle asset to go with the Square.
We don't have an easy way to generate simply rectangular sprites, this would assist with that.

pathunstrom:
Work to support rectangular shape assets.

Resolves #566

### DIFF
--- a/ppb/assets.py
+++ b/ppb/assets.py
@@ -1,3 +1,4 @@
+from ctypes import byref, c_int
 from typing import NamedTuple, Tuple, Union
 
 import sdl2.ext
@@ -10,6 +11,7 @@ from sdl2 import (
     SDL_DestroyRenderer,  # https://wiki.libsdl.org/SDL_DestroyRenderer
     SDL_SetRenderDrawColor,  # https://wiki.libsdl.org/SDL_SetRenderDrawColor
     SDL_RenderFillRect,  # https://wiki.libsdl.org/SDL_RenderFillRect
+    SDL_GetRendererOutputSize,  # https://wiki.libsdl.org/SDL_GetRendererOutputSize
 )
 
 from sdl2.sdlgfx import (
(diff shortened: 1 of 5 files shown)


AstraLuma (changes requested):
Can you add a viztest for this?

ppb/assets.py
```diff
@@ -92,33 +115,52 @@ def _draw_shape(self, renderer, rgb, **_):
         )
 
 
+def Square(r, g, b):
```
AstraLuma: Should we turn this into classes or just leave them as functions?

pathunstrom:
> Can you add a viztest for this?

Will do, I considered just, adding to our rectangle_sprite viz test, but I figure the example is also a visual test so can just. . . move it over.

ppb/assets.py
pathunstrom: Either way is fine with me.

I just wanted to keep this change backwards compatible and I want to document that it's basically just a wrapper of the "real" implementation.

ppb/assets.py
```diff
@@ -92,33 +115,56 @@ def _draw_shape(self, renderer, rgb, **_):
         )
 
 
+class Square(Rectangle):
+    """
+    A constructor for :class:`~ppb.Rectangle` that produces a square image.
+    """
+
+    def __init__(self, r, g, b):
+        super().__init__(r, g, b)
```
AstraLuma:
```suggestion
    def __init__(self, r, g, b):
        # This cuts out the aspect_ratio parameter
        super().__init__(r, g, b)
```

ppb/assets.py
```diff
@@ -92,33 +115,56 @@ def _draw_shape(self, renderer, rgb, **_):
...
             *rgb, 255,
             _check_error=lambda rv: rv < 0
         )
+
+
+class Circle(Ellipse):
+    """A convenience constructor for :class:`~ppb.Ellipse` that is a perfect circle."""
+
+    def __init__(self, r, g, b):
+        super().__init__(r, g, b)
```
AstraLuma:
```suggestion
    def __init__(self, r, g, b):
        # This cuts out the aspect_ratio parameter
        super().__init__(r, g, b)
```

AstraLuma (approved):
Looks good. I would strongly suggest incorporating the suggested comments, though.

Issue #482, opened 2020-05-24
Improve Asset loading Warning


pathunstrom:
The current warning is factual, but includes little details. This is confusing when using the default image (`...`) as if you're using the default squares and not adding an image file, it'll warn, but work correctly.

Should probably add a second sentence to the effect of "This may not be a problem, you can stop this warning by explicitly setting Sprite.image)" (See #479 for additional details.)

pathunstrom:
Closed with #584

PR #584, opened 2021-02-21
Improve Asset loading Warning Issue #482

bernardthered:


pathunstrom (approved):

AstraLuma (changes requested):
If I'm reading this correctly, this will add this warning text for every kind of asset with a `file_missing` behavior, but the text is only applicable to `ppb.Image`.

bernardthered:
@AstraLuma good catch! I corrected that issue, let me know what you think of the solution, I'd be happy to change anything about that :-)

bernardthered:
@pathunstrom it looks like another spurious unit test error (this time the error is "❗ Agent is not responding!"). Would you be able to press the rerun link when you get a chance?

ppb/assetlib.py
```diff
@@ -165,6 +165,10 @@ class AbstractAsset(abc.ABC):
     This defines the common interface for virtual assets, proxy assets, and
     real/file assets.
     """
+    # Can be overridden on child classes to add an additional message to show after "File not
+    # found: %r.". See Image.not_found_message for an example.
```
AstraLuma: This should be #: so it gets picked up by sphinx

AstraLuma (changes requested):
Minor changes. Otherwise, I'm happy with this design and implementation.

ppb/assetlib.py
bernardthered: I didn't know about that #: syntax, thanks for the pointer. I just read a bit about it and updated those two comment lines.

bernardthered:
I also updated the warning message to be more explicit, let me know if you think the original is better and I can change it back.

AstraLuma (approved):
bors r+

PR #611, opened 2021-05-16
changed `pos` to `position` on `Ship` class.

doobeh:
`Ship(pos=(...))` doesn't adjust the location. Changing to `position` to fix.  Doesn't cause an error, but might lead new users to think `pos` is a keyword to set position.


AstraLuma (approved):
Looks good! Just add yourself to the contributors list.

doobeh:
Added-- thanks!

Issue #384, opened 2020-02-28
Default Sprite Behaviors: KeyBoardMovementSprite


pathunstrom:
In the interest of improving the on-boarding ramp for new users and learners, and by extension making our live demonstrations a little more impressive, we'd like to include some Sprites with basic control features.

For this ticket we are defining a `KeyboardMovementSprite`. This sprite should allow you to move the object with four directional keys defined by the developer. The default should be the arrow keys, and we should define a module attribute for WASD movement. Other layouts will require definition from the end user.

We'd like to provide a collection of movement strategies: various kinds of linear interpolation and a basic physics model with a velocity, acceleration and max velocity.

This sprite will almost definitely need to be subclassed to add behaviors for the end user.

aaronmak:
I'll give this a go

sheenarbw:
@aaronmak  are you still working on this? I'm keen to get it done

PR #614, opened 2021-05-17
Add default KeyboardMovementSprite

aaronmak:
For https://github.com/ppb/pursuedpybear/issues/384

Create a KeyboardMovementSprite with sane defaults

Each key press would send the sprite in the appropriate direction with a
default deceleration

Also create basic classes for the user to change the keyboard direction
keys

### DIFF
--- a/ppb/features/default_sprites.py
+++ b/ppb/features/default_sprites.py
@@ -1,13 +1,14 @@
 """
 Theme: sprites with common default behaviours (motion)
 
-Types of motion include: relative to the motion of other sprites, moving 
-towards another object. 
+Types of motion include: relative to the motion of other sprites, moving
+towards another object.
 
 """
 
 import ppb
 import math
+from typing import Dict
 
 
 class TargetSprite(ppb.Sprite):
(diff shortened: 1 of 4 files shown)


ppb/features/default_sprites.py
```diff
@@ -58,3 +60,118 @@ def _exponential_decay(self, time_delta):
...
+    down: ppb.keycodes = ppb.keycodes.Down
+    right: ppb.keycodes = ppb.keycodes.Right
+
+
+@dataclass
+class WASDDirectionKeys(DirectionKeys):
+    up: ppb.keycodes = ppb.keycodes.W
+    left: ppb.keycodes = ppb.keycodes.A
+    down: ppb.keycodes = ppb.keycodes.S
+    right: ppb.keycodes = ppb.keycodes.D
```
AstraLuma: I'm trying to decide how I feel about using dataclasses for this--I'm feeling like not, since it's an additional thing to import.

AstraLuma (changes requested):
Hm. I think I see two problems with this:

1. It suffers from the diagonal speed problem--it is slightly faster to move diagonally than with an axis
2. If the play keysmashes, there doesn't seem to be an actual limit on speed.

@pathunstrom is more knowledgeable in how to improve this.

I do appreciate the inclusion of the easing function, will help a lot with the feel of it.

Some extra things we're going to ask this sprint (Doesn't have to be in the same PR):
* Docs (At least API docs)
* An example of using it

ppb/features/default_sprites.py
```diff
@@ -58,3 +60,118 @@ def _exponential_decay(self, time_delta):
...
+
+    :param direction_keys: Configuration for mapping direction to keys
+    :param direction: Starting direction
+    :param acceleration: The increase in speed for a repeated key press in the following 0.3 seconds.
+    :param friction: The decrease in speed when there are no key presses in the following 0.3 seconds.
+    :param initial_speed: The speed per second when the first of subsequent keypresses are made.
+    :param speed: The current speed per second.
+    :param max_speed: Maximum speed per second that the sprite can travel in a specific direction.
+    :param min_speed: Minimum speed per second that the sprite can travel in a specific direction.
+    :param last_key_pressed_at: The datetime of the last time any key was pressed.
```
AstraLuma: This isn't really a user-facing parameter.

ppb/features/default_sprites.py
```diff
@@ -58,3 +60,118 @@ def _exponential_decay(self, time_delta):
...
+        self.speed = speed * (1 + self.acceleration)
+
+
+    def slow_down(self, time_delta: float):
+        self.speed = self.speed * (1 - self.friction) ** time_delta
+
+
+    @property
+    def has_recent_keypress(self) -> bool:
+        return datetime.datetime.utcnow() - self.last_key_pressed_at <= datetime.timedelta(microseconds=300000)
```
AstraLuma: This is not the preferred way to do clocks in PPB.

* It's not monotonic
* It'll do funky things if the scene is paused

The preferred way is to sum [`time_delta`](https://ppb.readthedocs.io/en/stable/reference/events.html#ppb.events.Update.time_delta) on the update event.

If that is not an option (eg, in the animation framework), there's `ppb.utils.get_time()`, but that has trade-offs (namely, it keeps running, even if the scene is not the active one).

ppb/features/default_sprites.py
aaronmak: no worries, I'll remove this dependency

ppb/features/default_sprites.py
aaronmak: good point, will remove and make private

ppb/features/default_sprites.py
aaronmak: makes sense, will see what I can do

aaronmak:
> Hm. I think I see two problems with this:
> 
> 1. It suffers from the diagonal speed problem--it is slightly faster to move diagonally than with an axis
> 2. If the play keysmashes, there doesn't seem to be an actual limit on speed.
> 
> @pathunstrom is more knowledgeable in how to improve this.
> 
> I do appreciate the inclusion of the easing function, will help a lot with the feel of it.
> 
> Some extra things we're going to ask this sprint (Doesn't have to be in the same PR):
> 
> * Docs (At least API docs)
> * An example of using it

Thanks for this review! I've redid a bit how the speed is calculated using the Vector instead and it's length as the actual limit on speed. This should resolve both those issues I hope? So each keypress now "adds" a force in the pressed direction to the existing speed vector instead

aaronmak:
Happy to also add the docs and examples once this is more concrete. Sorry I'm a bit slow in replying because of the timezone difference

aaronmak:
I'm not sure why a test that seems unrelated is failing though

ppb/features/default_sprites.py
AstraLuma: That doesn't actually help. The problem wasn't the use of specifically `dataclasses` (which are in the stdlib and used elsewhere in the code). The problem is the ergonomics of using this kind of "object as a bag" in this context. In addition, by removing the use of `dataclass`, you have made it less usable by no longer having an `__init__`.

The problem is that in order to customize this, you have to import an additional class, for something that we generally accomplish with a dictionary:

```python
{
    keycodes.W: Vector(-1, 0),
    keycodes.Left: Vector(-1, 0),
    keycodes.D: Vector(-1, 0),
    keycodes.Right: Vector(-1, 0),
}
```

Which has two benefits:
* Less boilerplate
* More flexibility

tests/test_default_sprites.py
```diff
@@ -42,3 +42,117 @@ def test_target_sprite_min_speed():
     target_sprite.on_update(ppb.events.Update(1.0), lambda x: None)
 
     assert target_sprite.position.isclose((-1.2, -1.6))
+
+
+class TestWASDDirectionKeys:
+    def test_up(self):
+        WASDDirectionKeys().up == ppb.keycodes.W
```
AstraLuma: I just noticed, these tests don't seem to do anything?

AstraLuma (changes requested):

ppb/features/default_sprites.py
AstraLuma: To summarize: just because you're not using `dataclasses` doesn't mean you're not writing data classes.

aaronmak: I see what u mean now, I've used a dictionary instead

tests/test_default_sprites.py
aaronmak: yea sorry was gonna remove them

AstraLuma:
bors try

AstraLuma:
bors try

aaronmak:
close because it's very stale, not sure if it can be merged now

Issue #619, opened 2021-05-17
Occasional SdlMixerError for keyboard_and_mouse_controls example


a-hurst:
When running the keyboard_and_mouse_controls example, I almost always get an error saying that the Mixer audio device hasn't been opened. Here's the full traceback:

```
ahurst@d420:~/downloads/pursuedpybear/examples/keyboard_and_mouse_controls$ python3 targets.py
UserWarning: Using SDL2 binaries from pysdl2-dll 2.0.14
INFO:ppb.engine:Entering context
DEBUG:ppb.vfs:Opening laser1.ogg (__main__, laser1.ogg)
DEBUG:ppb.vfs:Opening target.png (__main__, target.png)
DEBUG:ppb.vfs:Opening target.png (__main__, target.png)
DEBUG:ppb.vfs:Opening target.png (__main__, target.png)
DEBUG:ppb.vfs:Opening target.png (__main__, target.png)
DEBUG:ppb.vfs:Opening player.png (__main__, player.png)
DEBUG:ppb.vfs:Opening target.png (__main__, target.png)
INFO:ppb.engine:Exiting context
Traceback (most recent call last):
  File "targets.py", line 90, in <module>
    log_level=logging.DEBUG,
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/__init__.py", line 125, in run
    eng.run()
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/engine.py", line 316, in run
    self.main_loop()
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/engine.py", line 345, in main_loop
    self.loop_once()
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/engine.py", line 361, in loop_once
    self.publish()
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/engine.py", line 387, in publish
    method(event, self.signal)
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/systems/sound.py", line 117, in on_play_sound
    chunk = event.sound.load()
  File "/home/ahurst/.local/lib/python3.7/site-packages/ppb/assetlib.py", line 221, in load
    return self._future.result(timeout)
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
[...]

PR #625, opened 2021-05-18
Fix sound

AstraLuma:
This will band-aid over #619.

However, it does not fix the underlying issue: Assets should only begin loading after all systems have been entered.


pathunstrom (approved):
We should put this at the top of the investigation cue in 1.1

bors r+

PR #627, opened 2021-05-18
Fix up some annotation issues

AstraLuma:


ppb/engine.py
```diff
@@ -42,6 +42,10 @@
 _cached_handler_names = {}
 
 
+Signal = Callable[[Any], None]
```
pathunstrom, line 45: ❤️

pathunstrom (approved):
bors r+

PR #630, opened 2021-05-30
Architecture Reference Documentation: Or How To Use PPB

pathunstrom:
Goal of this document is going to be addressing all the moving pieces of ppb, all of their relationships, in a prose style.

At the moment, this is just outlines and some thoughts on critical warnings or instructions to include. Open to all feedback.

### DIFF
--- a/docs/architecture.rst
+++ b/docs/architecture.rst
@@ -1,4 +1,4 @@
-How ppb Fits Together
+ How ppb Fits Together
 ===========================================================
 
 This document is an overall reference of what ppb is, how it's organized, and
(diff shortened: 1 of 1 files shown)


docs/architecture.rst
```diff
@@ -0,0 +1,154 @@
...
+3. Time?
+4. Camera
+
+Events
+-----------------------------------------------------------
+.. _arch-events:
+
+1. What is an event
+2. Event classes
+    1. dataclasses/attr classes
```
AstraLuma, line 71: reminder that using `dataclasses` isn't a requirement. it's just an easy way to make a data bag.

pathunstrom: That's as much a suggestion/how we do it as anything. But will keep in mind.

docs/architecture.rst
```diff
@@ -0,0 +1,189 @@
+ How ppb Fits Together
+===========================================================
+
+This document is an overall reference of what ppb is, how it's organized, and
+how to use the various pieces.
+
+The Conceptual Necessities
```
mxashlynn, line 7: This section is really well written -- I love it!
Only suggestion I have is to perhaps put a list of all the major topics to be covered right up front, something like:
- GameObject
- GameEngine
- Systems
- Scenes
- Events
- Assets

docs/architecture.rst
```diff
@@ -0,0 +1,189 @@
...
+-----------------------------------------------------------
+.. _arch-the-basics:
+
+Put some introduction here.
```
mxashlynn, line 11: I might suggest including something like:
"Games developed using PPB are generally made of **Sprites** organized into a hierarchy of **Scenes**.  These Sprites behave according to rules laid out in **Systems**, and communicate via **Events**.  Sprites load **Assets** from storage.  A special **GameObject** manages **Time** and another handles the virtual **Camera**."

docs/architecture.rst
```diff
@@ -0,0 +1,189 @@
...
+game time
+
+Input
+-----------------------------------------------------------
+
+.. _arch-input:
+
+Supported inputs
+what the flags look like
+reasoning for uncommon names (button.Primary, Secondary, Tertiary)
```
mxashlynn, line 187: I really like that this "reasoning for uncommon names" is included.
Doesn't need to be long, but this is the sort of "getting inside the library maintainer's head" thing I find super valuable!

docs/architecture.rst
mxashlynn: In thinking about it, a paragraph like this might mean the list I suggested above is not needed.

Issue #631, opened 2021-06-01
Write the tutorials


nilamo:
The Contributing doc mentions a Tutorials repo (https://github.com/ppb/pursuedpybear/blob/canon/CONTRIBUTING.md#i-want-to-teach), and the docs have a Tutorial page (https://ppb.readthedocs.io/en/stable/tutorials/index.html), but both are non-existent or empty.  Have the tutorials been removed?  Are the links broken?

Currently, all of the docs seem to be a Getting Started guide, and an API reference.

AstraLuma:
Nope, we just haven't written any yet.

nilamo:
Is that something I can take a shot at working on?  Are there specific tutorial ideas y'all have in mind?  Maybe feature-focused, like how to implement a pause menu, or an entire game from start to finish, like Pong?

Also, do you prefer raw markdown files, or jupyter notebooks?

pathunstrom:
We're open to both!

While it needs to eventually make it into the documentation somewhere, I recently wrote a blog post on this: https://ppb.dev/2021/05/27/help-1-0.html

The important notes:
* A tutorial is a "full program". You can use a full tutorial to teach a specific feature implementation, but it should include all of the steps to reach that point.
* For something like "here's a sample pause menu" that's for the cookbook/how to section. These are much smaller pieces of documentation focused _only_ on the feature being discussed.

As for works to be aware of, I've been (very slowly) writing a pong tutorial #461. We're open to all kinds of other games though!

We use sphinx to generate our documentation, which uses restructured text. If you want to provide a notebook that'd be great. Such a tutorial does not _currently_ have a home, so we'd have to have that discussion, plus what kinds of support we'd need available to make such a tutorial easier to use.

Also, if you want to go the route of jupyter notebooks, see previous art: https://opensource.com/article/20/5/python-games

None of this is to be discouraging, just consolidating all of the things that currently live in my head.

nilamo:
> None of this is to be discouraging

It absolutely is not.  With any semi-large project, it can be fairly daunting to know how (or even if) I can help, and the best ways to do so that fit in with the rest of the project.  Thanks for the links, I've got some reading to do later, now 😄

PR #632, opened 2021-06-01
Remove reference to tutorial project.

pathunstrom:
When we have a better idea of what our curriculum project looks like, we can re-add this section.


pathunstrom:
Removing due to confusion that helped spawn: #631

AstraLuma (approved):
bors r+

PR #644, opened 2021-10-07
Added failer to test_contexts to avoid CI timeout

### LINKED ISSUE #550: Add Failer to test_contexts
Relevant test: https://github.com/ppb/pursuedpybear/blob/canon/tests/test_engine.py#L69

We need to add a Failer with a reasonable timeout to this test to prevent infinite loops causing CI time outs.

Failer defaults to 1 second, which is reasonable for most tests that will fail with it.

Import `ppb.testutils.Failer` and add it to the basic_systems in the test in question.

sgavil:
Added `Failer` to basic systems in order to avoid timeouts. Closes #550


pathunstrom (approved):
This looks great. Thank you for your contribution!

bors r+

PR #645, opened 2021-10-08
Adds a second docs status badge for canon

pathunstrom:
Simple addition so we can determine if canon is also functioning.

(Alternatively, we can replace the stable badge with the canon badge since the canon one is arguably more important to us.)


AstraLuma (approved):
Sure, why not?

bors r+

PR #647, opened 2021-10-24
Fix #637

### LINKED ISSUE #637: Fix viztests/loading.py
`viztests/loading.py` stopped working because of an assetlib change (uses `__new__` instead of `__init__`).

Fix that.

AstraLuma:
Fixes #637


pathunstrom (approved):
since the failure is incidental:

bors r+

PR #651, opened 2021-12-20
Adds doc-utils pin for read the docs.

### LINKED ISSUE #650: Docs builds are failing: Resolve by updating doc-utils requirement
A change was made to doc-utils to replace traversal with a generator. Sphinx tries to reverse the generator, which fails.

The current solution is to pin doc-utils in requirements-docs.txt to <0.18 as documented here: https://github.com/sphinx-doc/sphinx/issues/9727

pathunstrom:
Resolves #650

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,4 +1,4 @@
 sphinx
 sphinx_rtd_theme
 importlib_metadata; python_version < "3.8"
-doc-utils<0.18
+doc-utils<0.18  # Fix for read the docs default sphinx version. If we update sphinx, test to see if we can remove this.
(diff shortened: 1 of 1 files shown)


AstraLuma:
Feel like there should be a comment about when to remove it?

pathunstrom:
I mean, I don't know the answer to "when to remove it" I assume it's tied to when we update sphinx?

AstraLuma:
That sounds like a good when.

It's a constraint tied to a bug, so just gotta make sure the next person knows that.

AstraLuma (approved):
bors r+

PR #654, opened 2021-12-21
Add helpful error when translate_point_to_screen fails due to AttributeError

### LINKED ISSUE #612: Add helpful error when translate_point_to_screen fails due to AttributeError
Example Error:

```
File "camera.py", line 163, in translate_point_to_screen
    return Vector(point.x - self.left, self.top - point.y) * self.pixel_ratio
AttributeError: 'tuple' object has no attribute 'x'
```

We don't necessarily want to force folks to use the ppb vector class, but we do expect the vector.x and vector.y to exist. That said, for people experiencing this error, they most likely set the position of a gameobject to a tuple or some other representation of a vector instead of a Vector.

We should capture this attribute error and raise a new one with a more helpful message. Encourage users to look at the Vector class and to set the position attribute using it instead of a tuple, list, or dictionary.

avarice-m:
Add detailed error info to camera.translate_point_to_screen(), which is helpful for tests

Add detailed error info to renderer.compute_rectangles(), which is helpful for when users use a VectorLike instead of a Vector

Fixes #612

### DIFF
--- a/ppb/camera.py
+++ b/ppb/camera.py
@@ -163,7 +163,11 @@ class Camera(RectangleShapeMixin, GameObject):
         :return: A vector in pixels.
         :rtype: Vector
         """
-        return Vector(point.x - self.left, self.top - point.y) * self.pixel_ratio
+
+        try:
+          return Vector(point.x - self.left, self.top - point.y) * self.pixel_ratio
+        except AttributeError as error:
+          raise TypeError("Expected Vector")
 
     def translate_point_to_game_space(self, point: Vector) -> Vector:
         """
--- a/ppb/camera.py
+++ b/ppb/camera.py
@@ -167,7 +167,7 @@ class Camera(RectangleShapeMixin, GameObject):
         try:
           return Vector(point.x - self.left, self.top - point.y) * self.pixel_ratio
         except AttributeError as error:
-          raise TypeError("Expected Vector")
+          raise TypeError(f"Expected Vector got {type(point).__name__}: {point!r}")
 
     def translate_point_to_game_space(self, point: Vector) -> Vector:
         """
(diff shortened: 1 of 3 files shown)


ppb/systems/renderer.py
```diff
@@ -311,7 +311,18 @@ def compute_rectangles(self, texture, game_object, camera):
...
+            center = camera.translate_point_to_screen(game_object.position)
+        except TypeError as error:
+            raise TypeError(f"""{type(game_object).__name__}.position was set to a tuple: 
+
+            (number, number)
+
+            Expected a vector:
+
+            Vector(number, number)
+            """)
```
pathunstrom:
```suggestion
            """) from error
```

Forgot this while we were pairing, this will also show users that the error originated in the camera.

ppb/camera.py
```diff
@@ -163,7 +163,11 @@ def translate_point_to_screen(self, point: Vector) -> Vector:
         :return: A vector in pixels.
         :rtype: Vector
         """
-        return Vector(point.x - self.left, self.top - point.y) * self.pixel_ratio
+
+        try:
+          return Vector(point.x - self.left, self.top - point.y) * self.pixel_ratio
+        except AttributeError as error:
+          raise TypeError("Expected Vector")
```
pathunstrom:
```suggestion
          raise TypeError(f"Expected Vector got {type(point).__name__}: {point}")
```

Got some feedback from @AstraLuma, this should cover it.

pathunstrom (changes requested):
Some suggestions for improved messaging. Otherwise good.

ppb/camera.py
AstraLuma: `{point!r}` to get the repr string

pathunstrom (approved):
Looks good.

PR #656, opened 2021-12-30
Fix the examples zip job

pathunstrom:
we were saving into `working_directory/examples.zip` and uploading `temporary_directory/examples.zip` This builds and uploads from the temp directory.


AstraLuma (approved):
honestly, as long as it's consistent

bors r+

PR #672, opened 2022-02-06
Import Warnings

### LINKED ISSUE #668: Import warnings in ppb __init__.py
We use the warnings module to tell users that their python version is deprecated. Unfortunately we seem to have forgotten to import the module!

Add `import warnings` under `import logging` so we don't error out while trying to warn our users!

jillslind:
This was the first time I used github to manage pull requests, so this was a bit of a challenge to wrap around how git actually operates (and, well, token generation). 

Fixes #668


AstraLuma (approved):
bors r+

PR #677, opened 2022-11-04
Actually distribute ppb.debug

AstraLuma:


setup.cfg
```diff
@@ -31,6 +31,7 @@ packages =
     ppb
     ppb.systems
     ppb.features
+    ppb.debug
```
nicoonoclaste, line 34: Is there a reason we shouldn't use setuptools' [autodiscovery], with `packages = find:`, to prevent similar issues in the future?

[autodiscovery]: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

AstraLuma: Not to my knowledge.

pathunstrom: Pretty much only if we were doing the "tests directory in the package" thing, but we're not so this is probably an okay solution.

Can we set up a test run as part of the release process that installs the package from outside the source (I'm okay with "from cirrus's artifacts") and run tests against the installed version?

Between the automating the package list and running tests against the wheel I think we prevent any more bad from happening.

PR #692, opened 2024-02-04
Fix rotation basis.

pathunstrom:
I think we set this to (0, -1) back before we flipped the the Y axis. It caused rotation to go backwards when using `Sprite.facing`


AstraLuma (approved):
The fix looks good, by can we get tests on this?

PR #700, opened 2024-04-08
set dimensions for missing image

### LINKED ISSUE #691: RectangleSprite objects create a Square image as default.
Probably need to special case our default asset creation to create an appropriately shaped rectangle instead of a square.

Sample code:

```python
import ppb


class Platform(ppb.RectangleSprite):
    pass


def setup(scene):
    scene.add(Platform(width=3, height=1, position=ppb.Vector(0, -5)))
```

We end up with a square of size 3 instead of a rectangle with a 3:1 aspect ratio.

image

What it should look like:

image

Noting, this currently gets set in the asset (https://github.com/ppb/pursuedpybear/blob/canon/ppb/systems/renderer.py#L93-L107) and would require some refactoring to make it possible to contextually choose the asset definition.

davidism:
When a RectangleSprite falls back to rendering a default shape instead
of an image, use its dimensions rather than a default square.

This adds a private `_file_missing_dimensions` tuple to `Image`, which defaults to `(1, 1)`. `RectangleSprite.__image__` sets this to its dimensions so that the default shape is a rectangle instead of a square.

fixes #691 

```
import ppb


class Platform(ppb.RectangleSprite):
    pass


class Player(ppb.Sprite):
    pass


def setup(scene):
    platform = Platform(width=3, height=1, position=ppb.Vector(0, -5))
    scene.add(platform)
    player = Player()
    scene.add(player)


ppb.run(setup)
```

Screenshot 2024-04-08 at 09 57 41

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -57,3 +57,4 @@ Halle Jones|HJones@aliacy.com||
 [Avarice M](https://github.com/avarice-m) | | |
 [Jami Schwarzwalder](https://github.com/jschwarzwalder) | |
 [Jill Slind](https://github.com/jillslind) | [jill.slind@gmail.com](jill.slind@gmail.com) | |
+[David Lord](https://github.com/davidism) | |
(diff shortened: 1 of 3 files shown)


pathunstrom (approved):
LGTM

PR #706, opened 2024-04-08
background colors breaking when added as floats

julielinx:
fixes issue #444


pathunstrom (approved):
This looks good, we'll figure out the test errors.

PR #712, opened 2024-05-20
fix diagonal movement rate

### LINKED ISSUE #709: KeyBoardMovementSprite moves too quickly when moving diagonally
The new KeyBoardMovementSprite handles up, down, left and right movements fine, but if you move diagonally it moves much faster.

jepler:
Note that `scale_to` gracefully handles the zero vector, so no special case is needed.

Jamie and I discussed and both had the gut feeling that leaving the `direction` property unchanged from now (so that its elements are still always +1, -1, or 0) was easier for implementation but could also make sense when explained.

Closes: #709


AstraLuma:
Where is Binnie?

AstraLuma (changes requested):
This code looks good. Would you like to add yourself to `CONTRIBUTORS.md`?

AstraLuma (approved):

Issue #449, opened 2020-04-26
Working with Colors


pathunstrom:
Currently, we expose color APIs pretty directly as RGB tuples. It'd be nice to support multiple color notations.

Personally, I like working with HSV, and have mostly been using an external tool to find my colors and convert them to RGB.

We should do a total pass at all the places we use color and see if we can support other notations.

* RGB 0-255
* HSV (There's at least two popular notations for this. Maybe pick one?)
* Hex codes

Open to other ideas.

ironfroggy:
Would it also be possible to support RGB 0.0-1.0 tuples?

pathunstrom:
It's not unreasonable.

I'm trying to decide if this is best as a Color interface and a bunch of concrete classes that allow the various top level implementations and output an (r, g, b) Tuple in [0, 255] for all of the low level SDL systems that need it or if it's one class that supports the various interfaces. . .

astronouth7303:
My primary concern is students: RGB appears in a lot of places, and most of those are 0-255.

PR #718, opened 2024-05-20
[WIP] Refactor colors

adamaaronson:
🚨 This is a breaking change, affecting the parameters when defining shapes and setting a scene's background color. 🚨 

Addresses issue #449.

## Done:

- Created `colors` module to represent RGB colors
- Refactored `Shape` colors, `Scene` background color, and `Text` color to use the new `Color` class
- Fixed example code and test code to use the new `Color` class
- Write tests for `colors`

## Next steps:

- Add more ways to define colors (by hex? by name?)

### DIFF
--- a/ppb/colors.py
+++ b/ppb/colors.py
@@ -20,9 +20,12 @@ class Color():
 
     @staticmethod
     def from_hsv(hue: float, saturation: float, value: float):
-        """Convert the given HSV color values to an RGB color."""
+        """
+        Create a Color from the given HSV color,
+        with hue, saturation, and value ranging from 0-1.
+        """
         red, green, blue = colorsys.hsv_to_rgb(hue, saturation, value)
-        return Color(int(red * 256), int(green * 256), int(blue * 256))
+        return Color(round(red * 255), round(green * 255), round(blue * 255))
 
 
 BLACK = Color(0, 0, 0)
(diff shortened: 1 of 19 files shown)


ppb/colors.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
+import colorsys
+
+@dataclass()
+class Color():
```
AstraLuma: Do we want to call this `RGB` instead?

ppb/colors.py
```diff
@@ -0,0 +1,36 @@
...
+            if value < 0:
+                raise ValueError(f'{key} cannot be less than 0.')
+            elif value > 255:
+                raise ValueError(f'{key} cannot be greater than 255.')
+    
+    def __iter__(self):
+        return (self.red, self.green, self.blue)
+
+    @staticmethod
+    def from_hsv(hue: float, saturation: float, value: float):
```
AstraLuma: I would rather have `HSV().to_rgb()`?

@pathunstrom your thoughts?

adamaaronson: Concerned about that approach - I think there should only be one class for color so that it is intuitive to users that that’s the class they should plug in whenever they need to use a color. Don’t want people to accidentally use an HSV color when the class is expecting an RGB color, for example.

adamaaronson: I decided to refactor it into an abstract base class `Color`, with `RGBColor` and `HSVColor` subclasses that all have a `to_rgb()` method that should generally be abstracted away from the user. Let me know what you think!

PR #726, opened 2026-05-18
Migrate to Actions

AstraLuma:
RIP Cirrus CI

Move everything to Actions


AstraLuma:
@pathunstrom i think this is good enough that we should merge it so we can unstick everything else

